PR #5, opened 2024-05-24
refactor: extract outlet context as a outside context

Innei:
For better typescript intelligence and maintenance


Innei:
I don't really recommend using getters and setters for state in OutletContext, I think I can use jotai or zustand to manage that later, I'll work on this

DIYgod:
> I don't really recommend using getters and setters for state in OutletContext, I think we can use jotai or zustand to manage that later, I'll work on this

Great! Thanks so much!

PR #7, opened 2024-05-25
fix: downgrade eslint to v8 let it works, and apply lint and fix some code

Innei:


Innei:
Set `eslint.codeActionsOnSave.rules` vscode's config, to ignore some eslint rules in auto-fix save action

PR #12, opened 2024-05-25
feat: dark mode

hyoban:
Supporting dark mode earlier can reduce the pain of late-stage adaptation and prevent us from writing hard-coded colors.

The colors of some locations may still need to be adjusted later.

ScreenShot 2024-05-25 23 57 42
ScreenShot 2024-05-25 23 58 28
ScreenShot 2024-05-26 00 03 23
ScreenShot 2024-05-25 23 59 14


DIYgod:
I think it's unnecessary to place the toggle button in such a valuable position on the homepage; putting it in Settings -> General might be a better approach.

hyoban:
Moved to settings page

image

PR #36, opened 2024-06-04
feat: subscription store and unread store, unread calcaulation logic

Innei:

Based on https://github.com/RSSNext/follow-server/pull/14

# Main Changes

Support unread store and subscription store


src/renderer/src/hooks/useUpdateEntry.tsx
```diff
@@ -1,4 +1,4 @@
-import type { EntriesResponse, ListResponse } from "@renderer/lib/types"
+import type { EntriesResponse, ListResponse } from "@renderer/models"
 import { Queries } from "@renderer/queries"
 import type { Response as SubscriptionsResponse } from "@renderer/queries/subscriptions"
```
DIYgod, line 3: `Response` does not exist

DIYgod (changes requested):

DIYgod (approved):
image

In my local scenario, some numbers are missing 🤯

PR #45, opened 2024-06-06
refactor: to api client

Innei:


src/renderer/src/lib/api-fetch.ts
```diff
@@ -0,0 +1,46 @@
...
+      if (options.body instanceof FormData) {
+        options.body.append("csrfToken", csrfToken)
+      } else {
+        (options.body as Record<string, unknown>).csrfToken = csrfToken
+      }
+    }
+  },
+})
+
+export const apiClient = hc<AppType>(import.meta.env.VITE_API_URL, {
```
hyoban, line 24: Looks like we can use `apiFetch` now

DIYgod (approved):

PR #73, opened 2024-06-19
refactor: feed store and struture route params

Innei:
This PR refactors the feedStore so that instead of logging the active state to the store, it now logs it to the URL, which makes use of the browser's forward/backward history stack and preserves the current page after a reload.

Also, optimized some re-render performance.

Due to a lot of refactoring, we need help testing some scenarios.

https://github.com/RSSNext/follow/assets/41265413/eb23fb8c-bde2-4947-9f3c-7479f37601a0


Innei:
before:
https://github.com/RSSNext/follow/assets/41265413/2d097867-a9ca-4ad3-bbe7-2c54ebf329b9

now:
https://github.com/RSSNext/follow/assets/41265413/516ff5a6-18d0-4dba-88ee-9743c140ac76

PR #89, opened 2024-06-25
feat: add discover feed form from rsshub

Innei:
# 🎉 Feature

More user-friendly RSSHub parameter input, import discovered feed from RSSHub

https://github.com/RSSNext/follow/assets/41265413/b4ec1c9f-198b-46f2-9c05-32628087c73a


src/renderer/src/modules/discover/types.ts
```diff
@@ -0,0 +1,33 @@
+export type RSSHubRouteType = Record<string, RSSHubRouteDeclaration>
+export interface RSSHubRouteDeclaration {
+  routes: Routes
+  name: string
+  url: string
+}
+type Routes = Record<string, RSSHubRoute>
+export type RSSHubRoute = {
+  path: string
```
TonyRL, line 9: path can be `string[]` as well. 
https://github.com/RSSNext/follow/blob/9301a186d60b76b3dfe57bd7eb127707b9d72da9/src/renderer/src/modules/discover/example-data.json#L896-L898

Innei: Yes, but we've done the processing at the server to prioritize the first item, so it's always string here

src/renderer/src/atoms/sidebar.ts
```diff
@@ -0,0 +1,11 @@
+import { createAtomHooks } from "@renderer/lib/jotai"
+import { atom } from "jotai"
+
+export const [
+  ,
```
zhangmo8, line 5: remove this

src/renderer/src/components/ui/markdown/plugins/container.ts
```diff
@@ -0,0 +1,31 @@
...
+  getMd: () => MarkdownIt,
+): ContainerArgs {
+  return [
+    container,
+    klass,
+    {
+      render(tokens, idx, _options, env: { references?: any }) {
+        const md = getMd()
+        const token = tokens[idx]
+        const info = token.info.trim().slice(klass.length).trim()
```
zhangmo8, line 19:
```suggestion
        const info = token.info.trim()?.slice(klass.length)?.trim()
```

Innei: This code was copied from vitepress, there are no problems here.

zhangmo8 (approved):
LGTM

DIYgod (approved):

PR #97, opened 2024-06-28
feat: local db

Innei:
## Why

To solve the problem of slow JSON persistence, this refactoring uses indexeddb to store data.
And we added a new option to enable or disable the use of offline data.

Fixed the problem of not being able to read star entries in offline mode.

## Performance

After testing 1680+ data items in Entry, it takes ~150ms (async time will not block ui) to reorganize all the data references and logic calculations.

This is at least 1x less than the JSON processing time.


src/renderer/src/store/entry/hooks.ts
```diff
@@ -27,7 +29,14 @@ export const useEntryIdsByFeedId = (feedId: string, filter?: EntryFilter) =>
         return result
       } else if (feedId === FEED_COLLECTION_LIST) {
         // TODO we can't filter collections in local mode by view
```
Innei, line 31:
```suggestion
```

DIYgod (approved):

PR #99, opened 2024-06-30
fix: import circular and claim modal

Innei:


src/renderer/src/queries/feed.ts
```diff
@@ -42,36 +47,40 @@ export const useFeed = ({ id, url }: { id?: string, url?: string }) =>
...
-      id: feedId!,
-    }).invalidateRoot()
-  },
-  async onError(err) {
-    toast.error(await getFetchErrorMessage(err))
-  },
-})
+    onSuccess() {
+      if (!feedId) return
+      Queries.entries
```
Songkeys, line 77: I think this would also cause circular import issue?

DIYgod (approved):

PR #132, opened 2024-07-11
feat: local search

Innei:
To implement local search, need to enable data persist and have cached data locally to use this feature.

Known issues:
Since the data is local, the data is not up-to-date. This may result in outdated results or finding resources that no longer exist on the server.

# Performance

Search in 1800+ entries with `content`

https://github.com/user-attachments/assets/a5be6812-8900-4c91-9f4e-4a9b4e3a5512


DIYgod:
I have trouble clicking to jump to the list

https://github.com/user-attachments/assets/50624a35-e899-4de4-9064-1a17b4031796

PR #152, opened 2024-07-23
build: remove uncessary files

Songkeys:
SCR-20240723-umqq

Only files selected as in screenshot are necessary?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -94,9 +94,21 @@ jobs:
         with:
           name: ${{ matrix.os }}
           path: |
-            out/make/**/*.dmg
+            out/make/**/*.zip
             out/make/**/*.exe
             out/make/**/*.AppImage
+            out/make/**/*.yml
+          retention-days: 90
+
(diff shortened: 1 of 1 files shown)


DIYgod:
zip and yaml files are required for automatic updates.

DIYgod:
> We'll keep the yml files then.
> 
> What about the zip file? Currently, only macos output has a zip file inside which has an `.app` file which is duplicated with the .dmg file bundled, resulting in a x2 artifect file size.

The electron-updater package uses zip files to download updates and explicitly prohibits the use of dmg files. I don't know the reason, but that's how it is.

https://github.com/electron-userland/electron-builder/blob/28c7864bc061dfcc077e1476686e496c33e2e4e5/packages/electron-updater/src/MacUpdater.ts#L89

Songkeys:
What do you think of the current change ⬆️. it separates the .dmg and .zip files for macos so that we don't downlaod a 2x size file.

DIYgod (approved):

PR #160, opened 2024-07-28
fix: incorrectly behavior when click radio label

lawvs:
This pull request fixes an issue where the radio label was not behaving correctly when clicked.

The issue was caused by incorrect usage of the `id` and `htmlFor` attribute in the `Radio` components.

This PR updates components to use the `useId` within `Radio` and respect the ID input by the user.

Before

https://github.com/user-attachments/assets/1c942d5c-a0e3-46ff-8081-d1deeba35704

After

https://github.com/user-attachments/assets/90d02eca-8a08-4908-9075-ceb8177c1953


Innei (approved):
Thank you!

PR #162, opened 2024-07-28
fix: button component with asChild

lawvs:
This pr uses `asChid` to remove unsafe ts-expect-error directive in the about page.

This pull request also fixes errors when using `asChild` in the button component.

Screenshot 2024-07-28 at 19 09 11

```
Error: React.Children.only expected to receive a single React element child.
```

The loading function also works normally.

Screenshot 2024-07-28 at 19 13 06

Refer to https://github.com/shadcn-ui/ui/issues/3117

### DIFF
--- a/src/renderer/src/components/ui/button/index.tsx
+++ b/src/renderer/src/components/ui/button/index.tsx
@@ -1,4 +1,4 @@
-import { Slot } from "@radix-ui/react-slot"
+import { Slot, Slottable } from "@radix-ui/react-slot"
 import { stopPropagation } from "@renderer/lib/dom"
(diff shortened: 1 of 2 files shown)


src/renderer/src/components/ui/button/index.tsx
```diff
@@ -107,6 +109,7 @@ export const ActionButton = React.forwardRef<
           <TooltipTrigger asChild>
             <Button
               as={as}
```
lawvs, line 111: I propose to remove the `as` property from the button and replace it with `asChild`.

Innei (approved):
LGTM

PR #170, opened 2024-07-31
style: reset the EntryReadHistory line-height

zhangmo8:
`text-lg` can affect `line-height`, causing avatar groups to be too close to the top
reset it with `leading-none`

Before: 
image

After: 
image


DIYgod:
Thanks!

PR #185, opened 2024-08-02
fix: allow toggle switch by clicking label

lawvs:
https://github.com/user-attachments/assets/40514114-b613-4204-bac7-4ae86253ee6a


Innei (approved):
Thank you!

PR #212, opened 2024-08-07
feat: pictures masonry

Innei:
Construct: masonry layout of the image.
Support: gird and masonry switching.

Following features:
Use 1:1 rendering when aspect is not present. And it will get the information of the image in the background and use the meta information to render in the next rendering.

https://github.com/user-attachments/assets/37536283-8c22-4936-8b7e-dd87fb807a3f

https://github.com/user-attachments/assets/cb417032-a343-4090-b744-3cf10f6500b1


src/renderer/src/modules/entry-column/Items/picture-masonry.tsx
```diff
@@ -0,0 +1,195 @@
...
+} & React.HtmlHTMLAttributes<HTMLDivElement> &
+PropsWithChildren) => {
+  const urls = useEntryStore((state) =>
+    state.flatMapEntries[entryId]?.entries?.media?.map((m) => m.url),
+  )
+
+  useEffect(() => {
+    if (!urls) return
+    for (const url of urls) {
+      fetchImageDimensions(url)
```
DIYgod, line 178: I recommend removing fetchImageDimensions, as the backend already handles this logic.

Additionally, I noticed that images with size information also request the meta API.

DIYgod (approved):

PR #219, opened 2024-08-13
fix: use official embed player in bilibili video when use web

lawvs:
This pull request fixes the issue where the embed player will check `window.REFERRER_LIST` by `PlayerUtil.isCertifiedReferrer()` and throw `[Permission] Not Allowed` error.


Innei (approved):
LGTM

PR #269, opened 2024-09-03
fix(discover): follow status change after modal closed

MaxtuneLee:
### Description

Use useQuery instead of useMutation to set a optimistic update after "add feed" modal closed, while useQuery is more reasonable since "search" is kind of fetching data not update data.

### Linked Issues
https://github.com/RSSNext/Follow/issues/254


hyoban:
You should not use `useQuery` directly in the component; see the `defineQuery` and `onSuccess` options in `FeedForm`.

MaxtuneLee:
> You should not use `useQuery` directly in the component; see the `defineQuery` and `onSuccess` options in `FeedForm`.

Thank for the reply, get it

Innei:
One way to look at it is that you don't need to use `useQuery` to drive the data update, you can just use `useSubscriptionByFeedId` hook to get whether the current feed is subscribed or not.

src/renderer/src/modules/discover/form.tsx
```diff
@@ -63,17 +63,6 @@ export function DiscoverForm({ type }: { type: string }) {
       keyword: prefix || "",
     },
   })
-  const mutation = useMutation({
```
iamvishalkhare, line 66: Hi there

MaxtuneLee:
> One way to look at it is that you don't need to use `useQuery` to drive the data update, you can just use `useSubscriptionByFeedId` hook to get whether the current feed is subscribed or not.

Thanks for the advice, I separated the Card from DiscoverForm into a separate FeedCard component, and use that hook to get latest status ,which seems to be working well now.

PR #273, opened 2024-09-05
feat: setting sync

Innei:
Link to #243


DIYgod:
It seems that only the general settings will be synchronized. Is this expected?

DIYgod (approved):

PR #279, opened 2024-09-06
fix(wallet): add missing space between words in wallet

cxheng315:
### Description
Add missing space between words

image

### DIFF
--- a/src/renderer/src/modules/settings/tabs/wallet/my-wallet-section/index.tsx
+++ b/src/renderer/src/modules/settings/tabs/wallet/my-wallet-section/index.tsx
@@ -40,7 +40,8 @@ export const MyWalletSection = () => {
       <div>
         <div className="text-sm">
           <p>
-            Power is an ERC-20 token on the 
+            Power is an ERC-20 token on the
+            {" "}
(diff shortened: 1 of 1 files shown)


src/renderer/src/modules/settings/tabs/wallet/my-wallet-section/index.tsx
```diff
@@ -40,7 +40,7 @@ export const MyWalletSection = () => {
       <div>
         <div className="text-sm">
           <p>
-            Power is an ERC-20 token on the
+            Power is an ERC-20 token on the
```
hyoban, line 43: This won't work, please use `{' '}` and make sure linter doesn't complain.

Innei: Do we switch to Preitter to format the code instead of eslint to keep the format uniform? @hyoban

hyoban: I'm OK with Prettier, you can easily switch to Prettier now.

PR #300, opened 2024-09-09
feat: custom feed title

### LINKED ISSUE #237: Custom feed title

hyoban:
### Linked Issues

Fixes #237

Fixes FOL-34

https://github.com/RSSNext/follow-server/pull/72

### DIFF
--- a/src/renderer/src/store/feed/store.ts
+++ b/src/renderer/src/store/feed/store.ts
@@ -21,11 +21,7 @@ class FeedActions {
     set({ feeds: {} })
   }
 
-  upsertMany(_feeds: FeedModel[]) {
-    const feeds = _feeds.map((feed) => ({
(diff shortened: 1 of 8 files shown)


src/renderer/src/store/feed/store.ts
```diff
@@ -20,7 +21,11 @@ class FeedActions {
     set({ feeds: {} })
   }
 
-  upsertMany(feeds: FeedModel[]) {
+  upsertMany(_feeds: FeedModel[]) {
+    const feeds = _feeds.map((feed) => ({
```
Innei, line 25: Feed Model data should be kept raw, so it is not recommended to override the data here.

Just use the subscription's override title in the UI rendering layer.

Innei (approved):

PR #315, opened 2024-09-10
fix: show episode cover for podcast

hyoban:
### Description

ScreenShot 2024-09-10 11 50 21

### Linked Issues

Fixes FOL-85
Fixes #309 
https://github.com/RSSNext/follow-server/pull/73

### Additional context

Why change the dependence version?

The version in the lock file is `8.0.0`, which does not match the version in the package.json.

PR #316, opened 2024-09-10
fix: wrong text wrap

hyoban:
### Linked Issues

https://discord.com/channels/1243823539426033696/1282819783691079734

### Additional context

Merge after #315


src/renderer/src/lib/utils.ts
```diff
@@ -240,3 +240,7 @@ export const getUrlIcon = (url: string, fallback?: boolean | undefined) => {
 }
 
 export const isEmptyObject = (obj: Record<string, any>) => Object.keys(obj).length === 0
+
+export const isTextWrappedInHtmlTag = (text: string) => {
```
Innei, line 244: Why not use CSS, `break-words`?

hyoban: Apply `break-words` to? When the content is text only, it seems `break-words` for prose do not work.

Innei: The `break-words` in tailwind are actually word-wrap, not word-break.

Try to use normal CSS for test

https://developer.mozilla.org/en-US/docs/Web/CSS/word-break

hyoban: Oh yes. Updated.

Innei (approved):

PR #321, opened 2024-09-10
feat: add .nvmrc config file

FatDoge:
### Description
Add ``.nvmrc`` config file.


FatDoge:
Different versions of Node.js has different corepack apis, add a .nvmrc to improve DX

PR #389, opened 2024-09-14
chore: add category & item hover:bg-native-active

Soya-xy:
### Description

QQ_1726290062540

When I delete a subscription or category on the left, I'm not sure if I want to delete the one item if I'm hovering between two items


Innei:
No. This goes against the UI design guidelines.

Issue #391, opened 2024-09-14
prompt the full list of categories when editing the feed


KarasuShin:
### Clear and concise description of the problem

when editing category of the feed, it should prompt the full list of categories, not only the ones in the current view. I hope that on the profile page, subscriptions from different views can be displayed under the same category. This requires them to use the same category name. However, when editing category information, the prompt list only shows categories available in the current view, so I have to re-enter the category names from other views. It would be ideal if all categories across views were listed from the start.

I think I can handle this and get a PR up soon

### Suggested solution

list all categories in the dropdown when editing the feed category.

### Validations

- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #392, opened 2024-09-14
feat: prompt the full categories when editing the feed

KarasuShin:
### Description

When editing the feed's category, the suggestion list only shows the categories from the current feed's view. If you want to use category names from other views, you have to re-enter them. This PR is intended to display categories from all views in the suggestions

### Linked Issues
#391 

### Additional context
before:
CleanShot 2024-09-14 at 14 45 55@2x

after:
CleanShot 2024-09-14 at 14 44 26@2x


Innei (approved):

Innei:
Thank you!

Issue #403, opened 2024-09-14
[i18n]: Added support for Japanese


upupnoah:
### Type of i18n contribution

New language support

### Language

Japanese

### Description

Added support for Japanese.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #404, opened 2024-09-14
feat(i18n): add Japanese language support

upupnoah:
Added Japanese translations to the application, including localization for app content, common phrases, external resources, language settings, and shortcuts.

Related to #403


Innei (approved):

PR #412, opened 2024-09-14
feat: show edit when self

ischanx:
### Description

change the text in the button

it should be "Edit" when i check the profile page of myself

before
image
after
image


Innei (approved):
Thank you!

PR #413, opened 2024-09-14
fix: read history always show

ischanx:
### Description

it is a bug from https://github.com/RSSNext/Follow/pull/377

it should be open when the mouse hover, but it always show now.

image


Innei (approved):

Innei:
Thank you1

PR #428, opened 2024-09-15
add features: de language support

jettpackx:
### Description

added de language support


locales/lang/de.json
```diff
@@ -0,0 +1,10 @@
+{
+  "langs.de": "Deutsch",
```
Innei, line 2: You should add `langs.de` in other lang json

jettpackx: Thanks for the suggestion! I've added `langs.de` to all the other language JSON files. :)

Innei:
Sorry, de support has been merged in, so this pr will be closed.

PR #433, opened 2024-09-15
fix: shortcuts kbd shake

ischanx:
### Description

click the button and all the row shake due to the border change

so i fixed the row height

20240915235649_rec_


Innei (commented):
Thank you!

PR #452, opened 2024-09-16
feat: support setting proxy for app

lawvs:
Fix https://discord.com/channels/1243823539426033696/1283713350995279883

- [x] main setting
- [x] setting ui in renderer

This pull request adds a new feature that enables users to set a proxy for the Follow app (only for `session.defaultSession`, not the entire app). This enhancement is particularly beneficial for users in regions with unreliable network conditions.

---

This PR still has many areas for improvement, such as a more user-friendly input interface, network test, support for other proxy methods, etc. But these parts can be optimized later based on user feedback.

In addition, this is a pure addition PR without any deletions :)

https://github.com/user-attachments/assets/a28d30b5-8a4f-4f05-950b-8a878b9b4829


apps/renderer/src/modules/settings/tabs/general.tsx
```diff
@@ -244,3 +250,22 @@ export const LanguageSelector = () => {
...
+      <SettingInput
+        type="text"
+        label={t("general.proxy")}
+        labelClassName="w-[150px]"
+        value={proxyConfig}
+        onChange={(event) => setProxyConfig(event.target.value.trim())}
+      />
+      <SettingDescription>{t("general.proxy.description")}</SettingDescription>
+    </SettingItemGroup>
+  )
```
lawvs, line 270: The proxy settings should not sync between clients, so we cannot use the `defineSettingItem` directly.

Innei: Another way to filter the synchronization status between multiple clients is to use `generalServerSyncWhiteListKeys` in `atoms/settings/general.ts`.

Innei: But, this implementation also sound

Innei (approved):

PR #463, opened 2024-09-17
fix: the filename case issue with generate-i18n-locale add zh-hk

jerryc127:
### Description

1. 在添加語言中發現，生成的文件名，如果以 - 分開，則 - 後面的字母需要大寫，語言文件才能生效，故推送此修改
2. 添加繁體中文（香港）

PixPin_2024-09-18_03-12-23

---

In the process of adding languages, I discovered that the generated filenames need to have uppercase letters after the hyphen (if the filename is separated by a hyphen) in order for the language files to work correctly. Therefore, I am submitting this modification.


Innei (approved):
Thank you

PR #478, opened 2024-09-18
feat(i18n): add a zh-cn README and improve the original README

Octopus058:
### Description

I delete duplicate content in README.md and add a Simplified Chinese Version (README.zh-cn.md).

There r some duplicate paragraphs in the original README, so i delete them and slightly modified the indentation of the first line (If it's not proper then restore is ok). Also, I think there's nothing better than adding a zh-CN README!

### Linked Issues

#487 

### Additional context

Think highly of you, hope your browser can achieve great success sincerely xD

Moreover, I have submitted Follow to HelloGithub Program!


Octopus058:
maybe shouldn't add space, i will correct it tonight.

DIYgod:
Sorry, this readme will change frequently, and we won't have enough resources to maintain multilingual documentation in the future.

Octopus058:
What about just delete the duplicate content in README?

DIYgod:
That makes sense, but I still want to emphasize those points upfront. 🤔

Octopus058:
ok i see, wish Follow to be better, cheers!🍺

PR #479, opened 2024-09-18
fix: allow prod and dev builds running in the same time

xc2:
### Description

Fixes the issue where the development build starts and immediately stops when the production build is running.

### Linked Issues

-


Innei (approved):
Thank you!

PR #511, opened 2024-09-19
feat: dual code theme

hyoban:


Innei:
This interaction seems a bit strange, maybe take a look at telegram's design interaction. Under the current theme, you can only set the code theme of the current theme.
CleanShot 2024-09-19 at 6  59 35@2x

hyoban:
Merged to one option

Innei (approved):

PR #517, opened 2024-09-19
feat: add support for pointer cursor

yuzhang9804:
### Description
- Add support for cursor pointer

image

### DIFF
--- a/apps/renderer/src/atoms/settings/ui.ts
+++ b/apps/renderer/src/atoms/settings/ui.ts
@@ -19,7 +19,7 @@ export const createDefaultSettings = (): UISettings => ({
   modalDraggable: true,
   modalOpaque: true,
   reduceMotion: false,
-  usePointerCursor: false,
+  usePointerCursor: null,
 
   // Font
   uiFontFamily: "SN Pro",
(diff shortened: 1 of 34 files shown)


Innei:
Even though I don't have this option turned on, the pointer to Switch is cursor.

yuzhang9804:
- The dependencies for useEffect haven't been added. 
- The cursor now has multiple CSS variables. Should this be adjusted? If there's no need for atomization, I feel one CSS variable might suffice.

jerryc127:
這種不是應該强制開啓的麽，不需要提供設置。
follow 現在很多可交互的item, 即沒有Hover 效果，也沒有 pointer cursor
很難察覺這是可以點擊的

---

This should be enabled by default and shouldn't require configuration. There are many interactive items in Follow, but they lack both hover effects and a pointer cursor, making it hard to tell that they are clickable.

yuzhang9804:
I've sent a direct message to Innei about this. The default cursor being set to 'default' is intended to make the application feel more like a native app.

Innei:
https://github.com/RSSNext/Follow/issues/442

yuzhang9804:
If there are plans to address this issue, should I close this PR for now?

PR #527, opened 2024-09-20
fix: fixed text overlap issues caused by style

zhangmo8:
### Description

fixed text overlap issues caused by style

before: 
image

after:
image


Innei (approved):
LGTM

PR #532, opened 2024-09-20
perf(electron): openElectronWindow

Simon-He95:


Innei (changes requested):
why?

Simon-He95:
There should be no need to judge the current environment every time you execute it. It can be determined when it is loaded for the first time, and then choose to use window or electron.

Innei:
Optimizations that do little to enhance, but instead degrade readability and code style

PR #542, opened 2024-09-21
refactor: deprecate hsl legacy syntax

xc2:
### Description

Follow mixes modern HSL syntax with legacy syntax. This can result in invalid values like `hsl(h, s, l / alpha)` or `hsl(h s l, alpha)` when using CSS variables and the Tailwind theme.

This PR deprecates all hsl legacy syntax from the codebase.

### Linked Issues

https://github.com/user-attachments/assets/0766ad2f-4f66-432a-a7a1-616c25938355


Innei (approved):
LGTM

PR #580, opened 2024-09-23
fix: use insertedAt as cursor for list

### LINKED ISSUE #578: 查看图片list报错
### Describe the bug

查看图片list报错，图片未读1500

### Feed Info

https://app.follow.is/list/60580187699502080?view=2

### Reproduction Video

CleanShot 2024-09-23 at 07 52 32@2x

### App Type

Electron / App

### Version

0.0.1-alpha.16

### Operating System

macOS

### Browser Info

_No response_

### Validations

- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.

### Contributions

- [ ] I am willing to submit a PR to fix this issue
- [ ] I am willing to submit a PR with failing tests (actually just go ahead and do it, thanks!)

hyoban:
### Linked Issues

fix #578

### DIFF
--- a/apps/renderer/src/modules/entry-column/hooks.ts
+++ b/apps/renderer/src/modules/entry-column/hooks.ts
@@ -8,7 +8,7 @@ import { useRouteParams, useRouteParamsSelector } from "~/hooks/biz/useRoutePara
 import { useAuthQuery } from "~/hooks/common"
 import { entries, useEntries } from "~/queries/entries"
 import { entryActions, useEntryIdsByFeedIdOrView } from "~/store/entry"
-import { useFolderFeedsByFeedId } from "~/store/subscription"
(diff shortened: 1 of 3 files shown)


apps/renderer/src/modules/entry-column/hooks.ts
```diff
@@ -71,6 +71,7 @@ export const useEntriesByView = ({
     view,
     ...(unreadOnly === true && { read: false }),
     isArchived,
+    isList: isSubscriptionAList(feedId),
```
Innei, line 74:
```suggestion
    isList: isSubscriptionList(feedId),
```
```

```

PR #608, opened 2024-09-24
Fix: English title display is incomplete

jerryc127:
### Description

image

image


Innei:
@jerryc127 Nice work, but the title and the following question are too loose

CleanShot 2024-09-24 at 10  07 37@2x

jerryc127:
> @jerryc127 Nice work, but the title and the following question are too loose
> 
> CleanShot 2024-09-24 at 10  07 37@2x

I change line-height: 1.25 to line-height: 1.15, it will be better

Innei:
I think it's still too loose here.

jerryc127:
image

我認爲已經是很好了，字母 g 已經很接近 ‘項目’兩個字了

PR #616, opened 2024-09-24
feat: add some i18n

ischanx:
### Description

image
image

### DIFF
--- a/apps/renderer/src/constants/shortcuts.ts
+++ b/apps/renderer/src/constants/shortcuts.ts
@@ -93,3 +93,11 @@ export const shortcuts = {
     },
   },
 } as const
+
+export const shortcutsType: { [key in keyof typeof shortcuts]: string } = {
+  feeds: "keys.type.feeds",
+  layout: "keys.type.layout",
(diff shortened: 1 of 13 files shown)


apps/renderer/src/constants/shortcuts.ts
```diff
@@ -93,3 +93,11 @@ export const shortcuts = {
     },
   },
 } as const
+
+export const shortcutsType: { [key in keyof typeof shortcuts]: string } = {
```
Innei, line 97:
```suggestion
export const shortcutsType: { [key in keyof typeof shortcuts]: I18nKeys } = {
```

Innei (changes requested):

apps/renderer/src/constants/shortcuts.ts
ischanx: Thanks for your review, but type I18nKeys can not be used, i add a new type I18nKeysForShortcuts

Innei (approved):

Innei:
Thank you

PR #632, opened 2024-09-25
style: fix the FloatMutedButton mispositioning

zhangmo8:
### Description

When the preview video is loaded for the first time, the mute icon flashes and misplaces

before:

https://github.com/user-attachments/assets/f80d9a36-084f-416e-a712-abb8689f04b8

after: 

https://github.com/user-attachments/assets/52f4d81b-1f15-46cf-b396-3a69f4414503


Innei:
We should hide mute button here. I add this logic

PR #646, opened 2024-09-26
fix(corner-player): pause playback when player finishes

HydenLiu:
### Description
* Before
image

* After
image


Innei (approved):
Thank you

PR #655, opened 2024-09-26
feat: view source content in app

lawvs:
Fix https://github.com/RSSNext/Follow/issues/247

This pull request introduces a new feature that allows users to view the source content in the follow app.

---

Warn banner for web

Screenshot 2024-09-28 at 15 37 44

Loading and transition

https://github.com/user-attachments/assets/76a9f0fe-dded-40fb-a3d3-07d28858181b

Modal in social and video

https://github.com/user-attachments/assets/ec97794e-0076-4639-83f2-8829d5646294

Screenshot 2024-09-28 at 15 40 51

In image view

https://github.com/user-attachments/assets/c0ed592a-3f6c-4576-bf2c-a24f3a98542d

---

~~Note: This is just a prototype PR designed to demonstrate the practicality and feasibility of this feature.~~

~~Before proceeding with further enhancements, I need some guidance on the following points:~~

~~1. Viewing the source in a modal is also a viable option. Should we adjust to view the source in a modal?~~
~~2. Currently, the feature is triggered by right-clicking the "Open in browser" btn. This is not an ideal solution, but I think the toolbar in the top-right corner is already crowded. Where do you think would be a better place for this button?~~


DIYgod:
1. I think the current implementation is better than modal, it gives me less pressure.

2. I think we can use the normal click event of "Open in browser", which means replacing the new window opening, after all, we already have another two `open in new window`: double-clicking the entry and clicking the entry title. And it would be better if we could set whether to view the source or open a new window in the settings, but by default, it should still view the source.

apps/renderer/src/modules/entry-content/index.tsx
```diff
@@ -174,88 +175,91 @@ export const EntryContentRender: Component<{
...
-        rootClassName={cn("h-0 min-w-0 grow overflow-y-auto @container", className)}
-        scrollbarClassName="mr-[1.5px]"
-        viewportClassName="p-5"
-        ref={scrollerRef}
-      >
-        <div
-          style={stableRenderStyle}
-          className="duration-200 ease-in-out animate-in fade-in slide-in-from-bottom-24 f-motion-reduce:fade-in-0 f-motion-reduce:slide-in-from-bottom-0"
-          key={entry.entries.id}
+      <div className="relative flex size-full flex-col overflow-hidden">
```
lawvs, line 178: Just wrap a new div here, try enabling "hide whitespace" to see the changes.

Innei:
That's great.

But have you considered adding a shortcut to this feature?

By the way, for the sake of UI consistency, we should use the Solid icon when the button is selected.

Innei:
And I've noticed a significant frame drop during the animation transition with the ViewSource open. Is there a way to optimize this.

Perhaps content rendering should be avoided during the transition, and only displayed after it is completed.

lawvs:
> And I've noticed a significant frame drop during the animation transition with the ViewSource open. Is there a way to optimize this.

I have optimized the transition when viewing the source code, and the frame drop issue has been somewhat alleviated.

https://github.com/user-attachments/assets/48f1f318-607c-4fab-bb04-0be6e06f5695

> But have you considered adding a shortcut to this feature?

Regarding the shortcut key, I think we can use cmd+b or ctrl+b (win). What do you think?

lawvs:
> But have you considered adding a shortcut to this feature?

> Regarding the shortcut key, I think we can use cmd+b or ctrl+b (win). What do you think?

After checking, I found that the shortcut cmd+B conflicts with the toggle feed sidebar, so it is not feasible. Additionally, assigning shortcuts without careful consideration of their functions is not a good practice. Perhaps we can consider it later.

Innei (approved):

Innei:
Nice work!

Issue #640, opened 2024-09-26
Follow cannot render the code block correctly, but Inoreader can render it.


cesaryuan:
### Describe the bug

There is an article in which Follow cannot render the code block correctly, but Inoreader can render it.

![image](https://github.com/user-attachments/assets/984f00c7-c945-470b-83de-33001f596389)

### Feed Info

https://seekiu.github.io/feeds/all.atom.xml

This feed item in it: [用 matplotlib 画出规范的论文插图](https://seekiu.github.io/2016/mpl-for-publish.html)

### App Type

Electron / App

### Version

Alpha.17

### Operating System

Windows 11

### Validations

- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.

### Contributions

- [ ] I am willing to submit a PR to fix this issue
- [ ] I am willing to submit a PR with failing tests (actually just go ahead and do it, thanks!)

hyoban:
![ScreenShot 2024-09-27 10 11 01](https://github.com/user-attachments/assets/9c7488c5-d634-4fdf-af47-b849c0308303)

PR #667, opened 2024-09-27
fix: null reference introduced in ab8c9e5

cesaryuan:
### Description

Fix the `Cannot use 'in' operator to search for 'type' in undefined` introduced in ab8c9e5.

Feed: https://dmesg.app/workers-function.html in https://dmesg.app/feed

image

### Linked Issues

#640


Innei (approved):

Innei:
The logic here doesn't seem quite right. I withdraw my request for approval

cesaryuan:
Maybe we should add some test cases to avoid such errors happening again.

PR #675, opened 2024-09-27
fix: add i18n in preview-media

kovsu:
### Description

CleanShot 2024-09-28 at 02 32 19@2x

CleanShot 2024-09-28 at 02 32 11@2x


DIYgod (approved):

DIYgod:
Thanks so much!

PR #692, opened 2024-09-28
feat(i18n): conditional display of 'browser' or 'new tab'

jerryc127:
### Description

feat: display 'browser' in app and 'new tab' in web

web
image

app
image


Innei (approved):
Thank you

PR #693, opened 2024-09-28
chore(transition): apply natural transition for avatar dropdown

konata33:
### Description

More natural animation effects when hovering over the mouse

**Old**
iShot_2024-09-28_19.55.39
**New**
iShot_2024-09-28_19.56.31

### DIFF
--- a/apps/renderer/src/modules/entry-content/components/EntryReadHistory.tsx
+++ b/apps/renderer/src/modules/entry-content/components/EntryReadHistory.tsx
@@ -72,8 +72,11 @@ export const EntryReadHistory: Component<{ entryId: string }> = ({ entryId }) =>
                 asChild
                 className={clsx(
                   "z-10 flex max-h-[300px] flex-col overflow-y-auto rounded-md border bg-background drop-shadow",
-                  // Animation, fade up
-                  "data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0 data-[state=open]:slide-in-from-bottom-3",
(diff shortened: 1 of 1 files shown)


apps/renderer/src/modules/entry-content/components/EntryReadHistory.tsx
```diff
@@ -99,14 +102,11 @@ const EntryUserRow: Component<{ userId: string }> = memo(({ userId }) => {
...
-    <li className="relative flex min-w-0 max-w-[50ch] rounded-md p-1 hover:bg-muted">
-      <button
-        type="button"
-        className="flex min-w-0 items-center gap-2 truncate"
-        onClick={() => {
-          presentUserProfile(userId)
-        }}
-      >
+    <li
+      onClick={() => presentUserProfile(userId)}
```
Innei, line 106: Why? `<li/>`'s role is not `button`, this will affect a11y

Innei (changes requested):

apps/renderer/src/modules/entry-content/components/EntryReadHistory.tsx
konata33: I see that it only works when clicking on the user name, not on the entire li tag. Perhaps I should adjust this back?

Innei (approved):

apps/renderer/src/modules/entry-content/components/EntryReadHistory.tsx
Innei: I solved this

Issue #701, opened 2024-09-29
When managing subscription feeds in the list, the input box is not cleared after adding a new feed.


Przeblysk:
### Clear and concise description of the problem

When managing subscription feeds in the list, the input box is not cleared after adding a new feed.

https://github.com/user-attachments/assets/ef44aed4-6041-4953-869f-11765c053d91

### Suggested solution

cleared after adding a new feed

### Validations

- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

SagarChandra07:
maybe this will help

https://mega.co.nz/#!qq4nATTK!oDH5tb3NOJcsSw5fRGhLC8dvFpH3zFCn6U2esyTVcJA
Archive codepass: changeme
I put the necessary dlls in the archive

PR #702, opened 2024-09-29
feat: cleared after adding a new feed

Przeblysk:
#701


hyoban:
Sorry, I did this at the same time as you. Next time you find a bug that can be fixed directly, you can directly submit a PR.

PR #705, opened 2024-09-29
feat(i18n): Chinese: webhook action setting

xuanxuan1231:
### Description

Chinese: webhook action setting


locales/settings/zh-CN.json
barnett617, line 1: maybe use Chinese word "网络钩子" like [wikipedia](https://zh.wikipedia.org/wiki/%E7%BD%91%E7%BB%9C%E9%92%A9%E5%AD%90)

locales/errors/zh-CN.json
barnett617, line 1: prefer to use a more friendly word

locales/settings/zh-CN.json
DIYgod: Oh no, I feel this is too strange, I might have to think for a long time to figure out what it really is.

xuanxuan1231: A chinese website called Gitee:
image

So can we use `Webhook` or `Webhooks`?

barnett617: Maybe just keep it "Webhooks" is good for now which is as same as it in en.json

locales/settings/zh-CN.json
```diff
@@ -39,6 +39,7 @@
   "actions.action_card.to": "到",
   "actions.action_card.translate_into": "翻译成",
   "actions.action_card.value": "值",
+  "actions.action_card.webhooks": "Webhook",
```
DIYgod, line 42: It seems that an 's' is missing, it's a countable noun.

PR #706, opened 2024-09-29
fix(i18n/zh): improve translation text in `wallet.transaction` table

at-wr:
### Description

Fixed the incorrect translation for “From” and “To” in `wallet.transaction`, for `zh_CN` and `zh_TW`.

### Linked Issues

None

### Additional context

None


barnett617 (commented):
disagree with this change, the current translation is better and more flexible

at-wr:
> disagree with this change, the current translation is better and more flexible

The original text "发送至" could be misunderstood as "Send to," which represents "接收者," while the English text "From" is obviously clearer. I believe it's unnecessary to use words like "至" or "自" because they can often confuse users. For example, I was actually confused the first time.

Innei (approved):

PR #708, opened 2024-09-29
fix: ShortcutMode cannot be closed using shortcuts keys

duowb:
Using `H` to open shortcut mode, pressing `H` again cannot close it


hyoban:
Thanks for your contribution!

PR #709, opened 2024-09-29
feat: Remember the open state of the category, Ensure that each View'…

Przeblysk:
feat: Remember the open state of the category, Ensure that each View's category is independent
close https://github.com/RSSNext/Follow/issues/602 https://github.com/RSSNext/Follow/issues/668


hyoban:
I can no longer see uncategorized feeds

Przeblysk:
> I can no longer see uncategorized feeds

😯 I hadn't thought of that I'll revise it.

Przeblysk:
> I can no longer see uncategorized feeds

I modified the display judgment, default display when feed is not grouped

https://github.com/RSSNext/Follow/pull/709/files#diff-59cf3a300e3e9c3e38aec7c41c219cfe98504302e861856f9799b1fd8c11b5d8R49

hyoban:
https://github.com/user-attachments/assets/88e0071c-83a5-4663-bc95-d1345b0cab69

I can't toggle the expanded state of automatic grouping by feed URL.

Przeblysk:
> ScreenShot.2024-09-29.14.51.08.mp4 
> I can't toggle the expanded state of automatic grouping by feed URL.

Okay, I'll fix it.

Przeblysk:
> ScreenShot.2024-09-29.14.51.08.mp4 
> I can't toggle the expanded state of automatic grouping by feed URL.


https://github.com/user-attachments/assets/2fbe771d-c460-4aab-a9c1-232bd72a4abe

I briefly tested it and it seems to be displaying fine

hyoban:
Should we save the open state to storage?

Przeblysk:
> Should we save the open state to storage?

If we want to store in localStorage, do we only read Storage to initialize the Store when the application is ready, and then only update Storage before the application exits?

hyoban:
LGTM. Let's merge it, we may not need to persist the state to storage

hyoban:
After adding a subscription, it does not appear in the list immediately and needs to be refreshed before it can be displayed. Can you help take a look? @Przeblysk

Przeblysk:
> After adding a subscription, it does not appear in the list immediately and needs to be refreshed before it can be displayed. Can you help take a look? @Przeblysk

Okay, I'll see what the problem is.

hyoban:
It should be OK after https://github.com/RSSNext/Follow/commit/b0a88312f52b2832d3b449aef15f5835b1cf3afc. Can you help verify it?

Przeblysk:
> After adding a subscription, it does not appear in the list immediately and needs to be refreshed before it can be displayed. Can you help take a look? @Przeblysk

https://github.com/Przeblysk/Follow/blob/dev/apps/renderer/src/modules/feed-column/category.tsx#L49

The state of categoryOpenStateData is lagging, causing the Effect of shouldOpen to be executed 2 times. It should probably be explicitly set to the open state instead of using toggle. i'll look into why the value of categoryOpenStateData is old.

Przeblysk:
> After adding a subscription, it does not appear in the list immediately and needs to be refreshed before it can be displayed. Can you help take a look? @Przeblysk

https://github.com/Przeblysk/Follow/blob/dev/apps/renderer/src/modules/feed-column/category.tsx#L49

The state of categoryOpenStateData is lagging, causing the Effect of shouldOpen to be executed 2 times. It should probably be explicitly set to the open state instead of using toggle. i'll look into why the value of categoryOpenStateData is

PR #710, opened 2024-09-29
fix: grid mode for images and video content page, shadow not fully displayed when the mouse hovers over the image

cuikaipeng:
In the grid browsing mode of images and on the video content display page, when the mouse hovers over an image, the shadow is not fully displayed and appears truncated at the top.
I identified that the issue was caused by the "inset-y-1" style in the 「ListItemHoverOverlay」 component.
As this is a foundational component, I opted not to modify it directly. Instead, I introduced an "overlayItemClassName" attribute to the 「EntryItemWrapper」 component to address the style issue.
iShot_2024-09-29_13 45 17


apps/renderer/src/modules/entry-column/item.tsx
```diff
@@ -55,10 +55,12 @@ function EntryItemImpl({ entry, view }: { entry: FlatEntryModel; view?: number }
     }
     case FeedViewType.Pictures: {
       Item = PictureItem
+      Item.overlayItemClassName = "top-0"
```
Innei, line 58: Don't change the Item's properties here, you should add them when the Item component is defined.

Innei (changes requested):

apps/renderer/src/modules/entry-column/item.tsx
cuikaipeng: oh, sorry, I don't know how to 「add them when the Item component is defined」
I defined a variable to set the "overlayItemClassName" property of 「EntryItemWrapper」
Is this ok?
iShot_2024-09-29_17 45 41

cuikaipeng:
I noticed a commit by the manager that refactored the component I had modified, so I closed this PR.

PR #717, opened 2024-09-30
feat(corner-player): support now playing

colmugx:
### Description

There is nothing in 'now playing', so we can use the `MediaSession API` to provide a more robust media playback experience.

Here is the effect when we used `MediaSession`

image

And something likes this when used the `Sketchybar`

image

### Additional context

There are some problems or troubles here:

- I'am not sure if this is the best place for this code.
- `cover processor`(I apologize for temporarily referring to it) is in `<FeedIcon />`, I suggest that we isolate this logic to more elegantly support `Now Playing`.


apps/renderer/src/modules/feed-column/corner-player.tsx
```diff
@@ -102,6 +108,20 @@ const CornerPlayerImpl = () => {
     scopes: HotKeyScopeMap.Home,
   })
 
+  useEffect(() => {
+    if (!entry || !feed) {
+      return noop
```
Innei, line 113: Why?

apps/renderer/src/modules/feed-column/corner-player.tsx
```diff
@@ -34,6 +36,9 @@ export const CornerPlayer = () => {
   const entry = useEntry(entryId)
   const feed = useFeedById(entry?.feedId)
 
+  //FIXME: I am not sure
```
Innei, line 39: Remove this

apps/renderer/src/atoms/media-session.ts
```diff
@@ -0,0 +1,18 @@
+import { atom } from "jotai"
+import { useEffect } from "react"
+
+import { createAtomHooks } from "~/lib/jotai"
+
+export const [, , useMediaMetadataValue, useMediaMetadataSetValue] = createAtomHooks(
```
Innei, line 6: It doesn't look like a global state manager is needed, so remove this

apps/renderer/src/modules/feed-column/corner-player.tsx
```diff
@@ -102,6 +108,20 @@ const CornerPlayerImpl = () => {
     scopes: HotKeyScopeMap.Home,
   })
 
+  useEffect(() => {
+    if (!entry || !feed) {
+      return noop
+    }
+
+    mediaMetadataSetValue(
```
Innei, line 116: Not suitable, can be put in `Player` management method

Innei (changes requested):

colmugx:
@Innei Absolutely! Yeah, it makes sense.

There's a document about [MediaSession](https://developer.mozilla.org/en-US/docs/Web/API/MediaSession). Besides metadata setup, it offers system media control handlers. It's the reason that I've structured this as an `atom` to support future updates.

But i removed it. and i wasn't sure if I understand all the logic of the player...

Innei (approved):
Nice implemention

PR #746, opened 2024-10-02
fix: resolve content blur after modal dragged

HydenLiu:
### Description
Fix two problems：
When the modal is moved
1、The content inside becomes blurred 
2、The button prompts are not displayed fully

https://github.com/user-attachments/assets/83dcfa68-2d49-4167-aedf-1ef6bc9b1763


Innei (approved):
Nice job

PR #761, opened 2024-10-05
fix: upon modifying the information in the Inbox, the list failed to update promptly

cuikaipeng:
### Description

on the Discover page, when modifying information in the Inbox, the list does not update immediately after the modification is completed.

https://github.com/user-attachments/assets/34d975b8-4cd7-4818-91b2-4d733c495c3f

### DIFF
--- a/apps/renderer/src/modules/discover/inbox-list-form.tsx
+++ b/apps/renderer/src/modules/discover/inbox-list-form.tsx
@@ -102,7 +102,14 @@ export function DiscoverInboxList() {
                     present({
                       title: t("sidebar.feed_actions.edit_inbox"),
                       content: ({ dismiss }) => (
-                        <InboxForm asWidget id={inbox.id} onSuccess={dismiss} />
+                        <InboxForm
+                          asWidget
+                          id={inbox.id}
+                          onSuccess={() => {
+                            inboxes.refetch()
+                            dismiss()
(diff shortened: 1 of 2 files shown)


apps/renderer/src/modules/discover/inbox-list-form.tsx
```diff
@@ -102,7 +102,14 @@ export function DiscoverInboxList() {
                     present({
                       title: t("sidebar.feed_actions.edit_inbox"),
                       content: ({ dismiss }) => (
-                        <InboxForm asWidget id={inbox.id} onSuccess={dismiss} />
+                        <InboxForm
+                          asWidget
+                          id={inbox.id}
+                          onSuccess={() => {
+                            inboxes.refetch()
```
Innei, line 109: Use optimistic update?

cuikaipeng: ok, I'll optimize it

Innei:
This implementation does not make sense. I will refactor later

cuikaipeng:
> This implementation does not make sense. I will refactor later

sorry to bother you, I would like to ask you for some advice. in useMutation, I handle optimistic updates through onMutate, is this “make sense“ ?   
code like ⬇️
iShot_2024-10-06_17 51 19

PR #771, opened 2024-10-05
refactor: make resizing more natural

xc2:
### Description

Resizing is not that natural now.

For example, when we move the top handle, the box becomes shorter from bottom.

https://github.com/user-attachments/assets/bda13586-200b-4809-a070-3ea179bf91a7

This PR combines `resizing` and `dragging` when the user is `resizing` from `top` or `left` to make sure the box is resizing following the cursor. Screencast:

https://github.com/user-attachments/assets/4d9c15b9-52c0-46e6-a9c7-918d5a9dd9e1

### DIFF
--- a/apps/renderer/src/components/ui/modal/stacked/modal.tsx
+++ b/apps/renderer/src/components/ui/modal/stacked/modal.tsx
@@ -109,12 +109,17 @@ export const ModalInternal = memo(
 
     const modalElementRef = useRef<HTMLDivElement>(null)
 
+    const dragController = useDragControls()
     const {
-      handlePointDown: handleResizeEnable,
+      handleResizeStop,
+      handleResizeStart,
+      relocateModal,
+      preferDragDir,
       isResizeable,
       resizeableStyle,
     } = useResizeableModal(modalElementRef, {
       enableResizeable: resizeable,
+      dragControls: dragController,
     })
     const animateController = useAnimationControls()
     useEffect(() => {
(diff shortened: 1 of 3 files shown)


apps/renderer/src/components/ui/modal/stacked/modal.tsx
```diff
@@ -346,7 +350,7 @@ export const ModalInternal = memo(
                   onClick={stopPropagation}
                   onSelect={handleSelectStart}
                   onKeyUp={handleDetectSelectEnd}
-                  drag
+                  drag={draggable && (preferDragDir || draggable)}
```
Innei, line 353:
```suggestion
                  drag={draggable && preferDragDir}
```

apps/renderer/src/modules/settings/modal/layout.tsx
```diff
@@ -80,7 +88,7 @@ export function SettingModalLayout(
         )}
         style={resizeableStyle}
         onContextMenu={preventDefault}
-        drag={draggable}
+        drag={draggable && (preferDragDir || draggable)}
```
Innei, line 91:
```suggestion
        drag={draggable && preferDragDir}
```

Innei (changes requested):

apps/renderer/src/modules/settings/modal/layout.tsx
xc2: it's in case of `draggable` is `"x"` or `"y"`.

`preferDragDir` is only available on resizing, when resizing is over, `preferDragDir` will be back to `null`.

So `draggable && (preferDragDir || draggable)` here means when the box is draggable, the direction is `preferDragDir` or value of `draggable`

PR #772, opened 2024-10-05
fix(styles): fix settings modal border style

sun0225SUN:
## before

image

## after

image

image


TonyRL:
#768

Innei (approved):
Thank you！

Innei:
But I think #768 makes sense.

Issue #778, opened 2024-10-06
Cannot login when developing in electron `pnpm run dev`.


cesaryuan:
### Describe the bug

I want to add a feature to Follow running in electron, so I ran `pnpm run dev`. But when I click `Log in with Github`, it will prompt that it cannot connect to `http://localhost:3000/auth/error`, I checked the ports that the process is listening on, and indeed there is no 3000.

![image](https://github.com/user-attachments/assets/39db90d3-f4f4-411c-853d-465d106174c0)

This is the log of `pnpm run dev`:
```
E:\desktop\Follow [fix-zero-size ≡]> pnpm run dev

> Follow@0.0.1-alpha.19 dev E:\desktop\Follow
> electron-vite dev

vite v5.4.7 building SSR bundle for development...
✓ 1840 modules transformed.
dist/main/multipart-parser-DuTErqOc.js      6.68 kB
dist/main/index-BrYzvZG-.js                48.01 kB
dist/main/index-Cbk4dfWk.js               379.89 kB
dist/main/index.js                      2,704.05 kB
✓ built in 3.68s

build the electron main process successfully

-----

vite v5.4.7 building SSR bundle for development...
✓ 2 modules transformed.
dist/preload/index.mjs  2.82 kB
✓ built in 10ms

build the electron preload files successfully

-----

dev server running for the electron renderer process at:

  ➜  Local:   http://localhost:5173/
  ➜  Network: use --host to expose

start electron app...


[main] env loaded: {
  VITE_WEB_URL: 'http://localhost:5173',
  VITE_API_URL: 'http://localhost:3000',
  VITE_IMGPROXY_URL: 'http://localhost:2873',
  VITE_INBOXES_EMAIL: '@follow.re'
}
10:08:59.256 (main) > http://localhost:5173
(node:2352) ExtensionLoadWarning: Warnings loading extension at C:\Users\cesar\AppData\Roaming\Follow (dev)\Follow\extensions\lmhkpmbekcpmknklioeibfkpmmfibljd:
  Permission 'notifications' is unknown or URL pattern is malformed.
  Permission 'contextMenus' is unknown or URL pattern is malformed.

(Use `electron --trace-warnings ...` to show where the warning was created)
Added Extension:  Redux DevTools
Added Extension:  React Developer Tools
10:09:03.188        > Follow: Next generation information browser https://github.com/RSSNext/follow
10:09:03.189        > Initialize Follow...
10:09:03.194        > migration took 5ms
[...]

PR #796, opened 2024-10-08
fix: improve developing experience with Electron on Windows

cesaryuan:
### Description

Currently, it is very inconvenient to develop with Electron on windows:
When developers run `pnpm run dev` in terminal, an electron instance will be started.
Then it alerts developer to login. After logging in in the opened browser, we are required to open `follow-dev://auth?token=xxx` in the browser to continue.

That's the problem. A separate electron instance will be started instead of using the already running one.
For the second instance, it does not have the environment variable needed to launch (including ELECTRON_RENDERER_URL and those in .env), which is loaded by the first instance but not passed to the second instance.

The essential question we need to address is why the second instance was initiated (why requestSingleInstanceLock not work).

So that's why. The second instance is launched by the system and not having the NODE_ENV configured to "development".
And it will make the second instance's appData directory to be `Follow` instead of `Follow (dev)`. Different appData directory means different instances.

So the solution is to check if the process is started with `follow-dev:` and set the NODE_ENV for it.

And now, we can develop with Electron on Windows very happily. 😁 (Remember set `VITE_API_URL` to `https://api.follow.is` in `.env` file)

### Linked Issues

#672, #778

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ You need to fill in the required environment variables first.
 cp .env.example .env
 ```
 
-Then run:
+Then set `VITE_API_URL` to `https://api.follow.is` and run:
 
 ```sh
 pnpm run dev
--- a/apps/main/src/init.ts
+++ b/apps/main/src/init.ts
@@ -16,7 +16,9 @@ const appFolder = {
   prod: "Follow",
   dev: "Follow (dev)",
 }
-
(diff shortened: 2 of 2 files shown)


Innei (approved):

zxhd863943427:
great! that's what I want! thank very much!

PR #800, opened 2024-10-08
feat: new user guide

### LINKED ISSUE #324: New user guide

hyoban:
### Description

https://github.com/user-attachments/assets/32d47e1e-843c-402e-bdff-ed214b151e13

### Linked Issues

close #324


apps/renderer/src/modules/new-user-guide/guide-modal-content.tsx
```diff
@@ -0,0 +1,208 @@
...
+        initial={false}
+        variants={{
+          inactive: {
+            backgroundColor: "var(--fo-background)",
+            borderColor: "hsl(var(--border) / 0.5)",
+            color: "#a3a3a3",
+          },
+          active: {
+            backgroundColor: "var(--fo-background)",
+            borderColor: "#ff760a", // theme-accent-500
```
Innei, line 167: Define a CSS variant for this then update tailwindcss config. e.g. `hsl(var(--a) / 1)`

apps/renderer/src/modules/new-user-guide/hooks.ts
```diff
@@ -0,0 +1,13 @@
...
+import { getStorageNS } from "~/lib/ns"
+
+export const [
+  ,
+  ,
+  useIsNewUserGuideFinished,
+  ,
+  getIsNewUserGuideFinished,
+  setIsNewUserGuideFinished,
+] = createAtomHooks(atomWithStorage(getStorageNS("is-new-user-guide-finished"), false))
```
Innei, line 13: Don't use local persist for this state, only new users need this guide, if old user changes this device or clear local data, the user guide should not be present.

So, you can read remote user settings, and just modify the `SettingSyncQueue` class to get whether the user has changed the settings (if there is a record on the server it means it was changed).

PR #808, opened 2024-10-08
feat: support preview feed view

lawvs:
Fix https://github.com/RSSNext/Follow/issues/430. This pull request introduces a new feature that allows users to preview feed views before following. 

https://github.com/user-attachments/assets/092f8453-5f79-46d9-a4a6-a9b41bd1153c

### DIFF
--- a/apps/renderer/src/modules/discover/feed-form.tsx
+++ b/apps/renderer/src/modules/discover/feed-form.tsx
@@ -350,7 +350,7 @@ const FeedInnerForm = ({
                 </div>
 
                 <ViewSelectorRadioGroup
-                  feedOrList={feed}
+                  feed={feed}
                   value={field.value}
                   {...form.register("view")}
                 />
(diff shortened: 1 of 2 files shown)


Innei:
Very good idea.

But the default expanding preview doesn't look good on it, which takes up a very large area.

lawvs:
> But the default expanding preview doesn't look good on it, which takes up a very large area.

Don't worry, we have several methods to alleviate this issue, such as placing it at the bottom or adding a collapse button for the preview item.


The following is my preferred solution， place the preview card on the bottom

image

hyoban:
We may need to create a simplified version of the component to preview feeds that have not yet been stored in the database. Now `EntryItem` needs to have `entry.id`.

lawvs:
We are currently waiting for @hyoban's new preview component, so this pull request is pending.

PR #828, opened 2024-10-09
fix: navigating from a tts player doesn't work properly

xc2:
### Description

This PR optimizes the navigation from the TTS player, includes:

- [x] fixes the problem, Follow always switch to the Audio view even the feed is not included in the Audio view.
- [x] supports inbox entry
- [x]  supports list entry

#### Before

https://github.com/user-attachments/assets/9447dc0d-6ed0-4f29-9382-fd1a2b6ac8c2

#### After

https://github.com/user-attachments/assets/468581d9-5851-437d-9e46-0b02304710e4

(Updated)

### DIFF
--- a/apps/renderer/src/modules/feed-column/corner-player.tsx
+++ b/apps/renderer/src/modules/feed-column/corner-player.tsx
@@ -24,6 +24,7 @@ import { FeedViewType } from "~/lib/enum"
 import { cn } from "~/lib/utils"
 import { useEntry } from "~/store/entry"
 import { useFeedById } from "~/store/feed"
+import { useListById } from "~/store/list"
 
 const handleClickPlay = () => {
   AudioPlayer.togglePlayAndPause()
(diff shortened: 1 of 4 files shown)


apps/renderer/src/modules/feed-column/corner-player.tsx
```diff
@@ -120,6 +121,23 @@ const CornerPlayerImpl = () => {
...
 
+  const navigateOptions = useMemo<NavigateEntryOptions>(() => {
+    if (!entry || !feed) return null
+    const options: NavigateEntryOptions = {
+      entryId: entry.entries.id,
+    }
+    if (feed.type === "inbox") {
+      options.inboxId = feed.id
+    } else if (feed.type === "list") {
+      options.inboxId = feed.id
```
xc2, line 132:
```suggestion
      options.listId = feed.id
```

PR #837, opened 2024-10-09
fix(shortcut): replace meta to ctrl in windows

jerryc127:


Innei (changes requested):
This implementation is a disaster for TS, it breaks type inference.

CleanShot 2024-10-10 at 11  52 12@2x

Currently, I have done Windows ctrl - meta conversion in kbd and ActionButton, native-menu. This PR implementation is not necessary

PR #822, opened 2024-10-09
feat: add CopyButton for url Tooltip of the text content

cuikaipeng:
### Description

add CopyButton for url Tooltip of the text content
sometimes need to copy this url to send to others, so I made this

**Before**
iShot_2024-10-09_14 04 09

**After**
iShot_2024-10-09_14 05 13

PR #845, opened 2024-10-10
refactor: simplify context menu handling and add options

cesaryuan:
### Description

For a long time, there have been several bugs in the right-click menu that have persisted. However, due to their lack of obviousness, they have gone unnoticed. Recently, the inability of the long-used context menu to appear (introduced by 5e91f19) has brought this issue to my attention. 

This problem is caused by a conflict between `Menu.popup` called in `apps\main\src\lib\context-menu.ts:registerContextMenu` and `Menu.popup` called in `"electron-context-menu"` used in `apps\main\src\init.ts`, resulting in the menu not being displayed. And after taking a look, I found that these two pieces of code are completely redundant and should be merged into one place. 

**Therefore, the issues solved by this PR are as follows:**
1. The problem that the context menu of the input box cannot pop up (isEditable).
2. The problem that the context menu at the link cannot pop up (linkURL).
3. The problem that the context menu after selecting text cannot pop up (selectionText).
4. The problem of code redundancy.

### Linked Issues

#822

### Additional context

**Before:**

https://github.com/user-attachments/assets/bca299f2-69ee-4e0b-8371-e406c682a419

**After:**

https://github.com/user-attachments/assets/e7f428a6-90af-4696-84fa-91a1f4bf4ef1


Innei (approved):
Nice fix!

PR #848, opened 2024-10-10
fix: Mouse click may briefly overlap with background elements

### LINKED ISSUE #838: [BUG] Mouse Click May Briefly Overlap with Background Elements After Returning to Subscription Input Box from Preview
### Describe the bug
点击预览后返回订阅源输入框时，鼠标点击可能短暂穿透当前输入框，选中后台页面的其他添加选项

https://github.com/user-attachments/assets/afa3b6b9-4553-469e-ba43-091d712564c0

### Feed Info

X

### Reproduction Video

_No response_

### App Type

Electron / App

### Version

v0.0.1-alpha.19

### Operating System

Windows 11

### Browser Info

_No response_

### Validations

- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.

### Contributions

- [ ] I am willing to submit a PR to fix this issue
- [ ] I am willing to submit a PR with failing tests (actually just go ahead and do it, thanks!)

Przeblysk:
close #838
The reason for this is that when multiple modals were being displayed off-rotation, the old modal had pointer-events-none set so that clicks were not blocked.

### DIFF
--- a/apps/renderer/src/components/ui/modal/stacked/overlay.tsx
+++ b/apps/renderer/src/components/ui/modal/stacked/overlay.tsx
@@ -24,7 +24,7 @@ export const ModalOverlay = forwardRef(
         ref={ref}
         id="modal-overlay"
         className={cn(
-          "pointer-events-none fixed inset-0 z-[11] rounded-[var(--fo-window-radius)] bg-zinc-50/80 dark:bg-neutral-900/80",
+          "fixed inset-0 z-[11] rounded-[var(--fo-window-radius)] bg-zinc-50/80 dark:bg-neutral-900/80",
           blur && "backdrop-blur-sm",
           className,
         )}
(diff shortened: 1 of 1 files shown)


Innei:
@Przeblysk This fix is wrong. I will revert this change in the next commit.

The reason for this is that the modal overlay must be ignoring mouse events, so that clicking on an element underneath the modal can interact with it during the transition state of the modal closing.
The problem with issue is not the overlay, but the addition of document.body point-event: none to the modal.

Przeblysk:
@Innei That way the #838 problem will still be there

https://github.com/user-attachments/assets/42c79ee5-df35-42ce-a03b-882a3d64df90

PR #854, opened 2024-10-10
fix: cmdk panel layout

Debbl:
### Description
fix `cmdk` panel footer layout.

image

fixed the `ScrollArea` area and changed some spacing.

image

result

image

### DIFF
--- a/apps/renderer/src/modules/panel/cmdk.tsx
+++ b/apps/renderer/src/modules/panel/cmdk.tsx
@@ -149,71 +149,76 @@ export const SearchCmdK: React.FC = () => {
         />
         <div className={cn(styles["status-bar"], isPending && styles["loading"])} />
 
-        <ScrollArea.ScrollArea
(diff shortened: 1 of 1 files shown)


Innei:
I can't reproduce.

This is the current UI and there is no problem.
CleanShot 2024-10-10 at 3  16 02@2x

Debbl:
> I can't reproduce.
> 
> This is the current UI and there is no problem. CleanShot 2024-10-10 at 3  16 02@2x

Try scrolling the search results across the footer.

image

Innei:
CleanShot 2024-10-10 at 3  41 41@2x

Debbl:
@Innei this is my Chrome version `129.0.6668.101`，If there is still a problem with the same version, it should be a problem with my device, I haven't tested it on other devices.

image

db1d19076c781696b9421f53dd79c35a6f49a324
image

Innei (approved):

PR #855, opened 2024-10-10
refactor(dock): simplify badge count logic

kovsu:


Innei (approved):
I don't think it's necessary to show `99+` here, it makes more sense to show the actual number.

Przeblysk:
Agree that it makes more sense to show the actual numbers

PR #884, opened 2024-10-11
feat: support use default browser open source content

BlackHole1:
### Description

Support adding an option in the settings page to allow users to choose whether to open the original content through the default browser.

### Additional context

Close: https://discord.com/channels/1243823539426033696/1294094455485366312

### DIFF
--- a/apps/renderer/src/atoms/settings/general.ts
+++ b/apps/renderer/src/atoms/settings/general.ts
@@ -7,6 +7,7 @@ import { createSettingAtom } from "./helper"
 const createDefaultSettings = (): GeneralSettings => ({
   // App
   appLaunchOnStartup: false,
+  useDefaultBrowserOpenSourceContent: false,
   language: "en",
   // Data control
   dataPersist: true,
@@ -42,6 +43,7 @@ export const subscribeShouldUseIndexedDB = (callback: (value: boolean) => void) 
 
 export const generalServerSyncWhiteListKeys: (keyof GeneralSettings)[] = [
   "appLaunchOnStartup",
+  "useDefaultBrowserOpenSourceContent",
   "dataPersist",
   "sendAnonymousData",
   "language",
--- a/apps/renderer/src/hooks/biz/useEntryActions.tsx
+++ b/apps/renderer/src/hooks/biz/useEntryActions.tsx
@@ -13,6 +13,7 @@ import {
   setReadabilityContent,
   setReadabilityStatus,
 } from "~/atoms/readability"
+import { useGeneralSettingKey } from "~/atoms/settings/general"
 import { useIntegrationSettingKey } from "~/atoms/settings/integration"
 import {
   setShowSourceContent,
@@ -166,6 +167,9 @@ export const useEntryActions = ({
   const enableInstapaper = useIntegrationSettingKey("enableInstapaper")
   const instapaperUsername = useIntegrationSettingKey("instapaperUsername")
   const instapaperPassword = useIntegrationSettingKey("instapaperPassword")
+  const useDefaultBrowserOpenSourceContent = useGeneralSettingKey(
+    "useDefaultBrowserOpenSourceContent",
+  )
 
   const checkEagle = useQuery({
     queryKey: ["check-eagle"],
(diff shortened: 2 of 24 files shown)


apps/renderer/src/atoms/settings/general.ts
```diff
@@ -42,6 +43,7 @@ export const subscribeShouldUseIndexedDB = (callback: (value: boolean) => void)
 
 export const generalServerSyncWhiteListKeys: (keyof GeneralSettings)[] = [
   "appLaunchOnStartup",
+  "useDefaultBrowserOpenSourceContent",
```
Innei, line 46: Remove this, This item can be sync to server

apps/renderer/src/atoms/settings/general.ts
```diff
@@ -7,6 +7,7 @@ import { createSettingAtom } from "./helper"
 const createDefaultSettings = (): GeneralSettings => ({
   // App
   appLaunchOnStartup: false,
+  useDefaultBrowserOpenSourceContent: false,
```
Innei, line 10: Move to bottom, This item should not be categorized as an App

apps/renderer/src/hooks/biz/useEntryActions.tsx
```diff
@@ -166,6 +167,9 @@ export const useEntryActions = ({
   const enableInstapaper = useIntegrationSettingKey("enableInstapaper")
   const instapaperUsername = useIntegrationSettingKey("instapaperUsername")
   const instapaperPassword = useIntegrationSettingKey("instapaperPassword")
+  const useDefaultBrowserOpenSourceContent = useGeneralSettingKey(
```
Innei, line 170: There is no need to observe to this setting item.

Innei: Remove this code

apps/renderer/src/hooks/biz/useEntryActions.tsx
```diff
@@ -396,6 +400,12 @@ export const useEntryActions = ({
         active: showSourceContent,
         onClick: () => {
           if (!populatedEntry.entries.url) return
+
+          if (useDefaultBrowserOpenSourceContent) {
```
Innei, line 404:
```suggestion
          if (getGeneralSettings().useDefaultBrowserOpenSourceContent) {
```

apps/renderer/src/hooks/biz/useEntryActions.tsx
```diff
@@ -480,6 +490,7 @@ export const useEntryActions = ({
     showSourceContentModal,
     read,
     unread,
+    useDefaultBrowserOpenSourceContent,
```
Innei, line 493: Remove this code

apps/renderer/src/modules/settings/tabs/general.tsx
```diff
@@ -53,6 +53,10 @@ export const SettingGeneral = () => {
     setGeneralSetting("appLaunchOnStartup", checked)
   }, [])
 
+  const saveSourceContentSetting = useCallback((checked: boolean) => {
```
Innei, line 56: This `useCallback` is unnecessary, remove this wrap or just write function inline

Innei (changes requested):

apps/renderer/src/atoms/settings/general.ts
Innei: And this name is too lang

BlackHole1: > Move to bottom, This item should not be categorized as an App

Which category do you recommend placing it in? Perhaps we should add a new category? Such as _interaction_ or _misc_.

> And this name is too long.

What do you think of `openContentInBrowser`?

apps/renderer/src/modules/settings/tabs/general.tsx
```diff
@@ -72,6 +76,13 @@ export const SettingGeneral = () => {
             },
           }),
           LanguageSelector,
+          defineSettingItem("useDefaultBrowserOpenSourceContent", {
+            label: t("general.use_default_browser_open_source_content"),
+            disabled: !tipcClient,
```
Innei, line 81: Remove this

apps/renderer/src/hooks/biz/useEntryActions.tsx
Innei: The logic here is incorrect. You need to make a choice with `openInBrowser` above

Innei (changes requested):

BlackHole1:
> You need to make a choice with openInBrowser above

I don't quite understand the meaning of this sentence.

BTW. I think this feature might not be suitable, because currently, users can click the article's title or double-click the item to open the browser. 

There are already two entry points provided to the users, so there's no need to add another one.

Therefore, I will close this PR. If anyone has different opinions, feel free to comment :)

PR #891, opened 2024-10-12
feat: add loading when Discover Inbox querying data

cuikaipeng:
### Description

refered to the loading style of the RSS3, User, and Transform modules in Discover


hyoban:
We can start showing the loading status from `TableBody`

cuikaipeng:
> We can start showing the loading status from `TableBody`

ok

PR #899, opened 2024-10-12
fix: app icon incorrect in Windows Control Panel

BlackHole1:
### Description

Now: 
CleanShot 2024-10-12 at 14 28 26@2x

Before:
图片

### Additional context

Close: https://discord.com/channels/1243823539426033696/1294232031148118088

### DIFF
--- a/forge.config.ts
+++ b/forge.config.ts
@@ -154,6 +154,7 @@ const config: ForgeConfig = {
     new MakerSquirrel({
       name: "Follow",
       setupIcon: "resources/icon.ico",
+      iconUrl: "https://app.follow.is/favicon.ico",
     }),
     new MakerAppImage({
       options: {


forge.config.ts
```diff
@@ -154,6 +154,7 @@ const config: ForgeConfig = {
     new MakerSquirrel({
       name: "Follow",
       setupIcon: "resources/icon.ico",
+      iconUrl: "https://app.follow.is/favicon.ico",
```
Innei, line 157: Can URl use local paths?

BlackHole1: CleanShot 2024-10-12 at 16 37 40@2x

Ref: https://github.com/electron/windows-installer/issues/151

The configuration options for @electron-forge/maker-squirrel can be found at https://github.com/electron/windows-installer.

Previously, the nupkg restricted iconUrl to be an http(s) address. The nupkg team claimed to have fixed this issue in 2019, but according to community reports, it doesn't work on Windows 10.

Testing shows that using: `file://${__dirname}/resources/icon.ico` can work (the downside might be that it doesn't display correctly on Windows 10).

BlackHole1: According to the latest nupkg documentation, iconUrl has been deprecated. The current recommendation is to use `<icon>`, and this tag supports local image resources. Unfortunately, the built-in nupkg version in windows-installer is 2.8.5, and this tag is only supported from nupkg@5.3.0. Additionally, windows-installer does not provide any method to specify the path of nupkg.

BlackHole1: BTW. Why not use electron-builder? Forge currently only supports MSI/Squirrel on Windows. However, electron-builder supports NSIS. Using NSIS won't have this issue,

Innei (approved):

PR #940, opened 2024-10-14
Refactor useFeedActions component to conditionally add a separator

jerryc127:
### Description

Refactor useFeedActions component to conditionally add a separator in the menu list based on the length of listByView array.

Refactor FeedCategoryImpl component to conditionally add a separator in the menu list based on the length of listList array.

### DIFF
--- a/apps/renderer/src/hooks/biz/useFeedActions.tsx
+++ b/apps/renderer/src/hooks/biz/useFeedActions.tsx
@@ -115,9 +115,7 @@ export const useFeedActions = ({
               },
             }
           }),
-          {
-            type: "separator",
-          },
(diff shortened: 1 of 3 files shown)


apps/renderer/src/hooks/biz/useFeedActions.tsx
```diff
@@ -115,9 +115,7 @@ export const useFeedActions = ({
               },
             }
           }),
-          {
-            type: "separator",
-          },
+          ...(listByView.length > 0 ? [{ type: "separator" as const }] : []),
```
Innei, line 118:
```suggestion
          listByView.length > 0 && { type: "separator" as const }
```

Innei (changes requested):

Innei (approved):

PR #967, opened 2024-10-15
fix: adjust position limit for drawer modal

xhuajin:
### Description

When using the desktop version (Windows 10), I noticed that when the ‘Trending’ section is opened, the ‘Minimize, Reduce, Close’ buttons in the top right corner overlap the ‘Trending’ modal, obscuring about a third of the ‘Trending’ close button.

bug screenshot:
image

Upon inspection, it was found that the z-index of the three buttons in the top right corner was set to 9999, whereas the z-index of the ‘Trending’ modal was only 1001. After attempting to change the z-index of the three buttons from 9999 to 1000, everything appeared to function normally. However, to minimize the impact, this PR opts to resolve the issue by increasing the z-index of the modal from 1001 to 10000.

Screenshot after repair:
image


lawvs:
Simply increasing the z-index value isn't the best solution. It's generally expected for the application's action buttons to remain on the top layer.

Innei:
This implementation doesn't make sense, you should adjust the boundaries of the modal instead of the Z-index.

xhuajin:
According to your suggestion, I remove the previous method of modifying z-index and add 'top: ${ElECTRON_CUSTOM_TITLEBAR_HEIGHT}px' to solve this problem.

Now the trending modal will look like:
image

lawvs (approved):
Nice works!

xhuajin:
Got it. Thanks for your reply

Issue #959, opened 2024-10-15
Image preview's blurhash transition state with extra blank at the bottom modal


Innei:
### Describe the bug

When the image is not loaded, there is a gap at the bottom of the blurhash. And the blurhash disappears directly after the transition state is complete, without an overlay transition with the image's appearance.

### Feed Info

N/A

### Reproduction Video

https://github.com/user-attachments/assets/6d211eb9-2f86-443b-adf8-3dc0b75c30cc
![CleanShot 2024-10-15 at 7  33 24@2x](https://github.com/user-attachments/assets/b4060a27-f059-43ff-9d7b-0d24b1a5d840)

### App Type

Web

### Version

N/A

### Operating System

N/A

### Validations

- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.

### Contributions

- [ ] I am willing to submit a PR to fix this issue
- [ ] I am willing to submit a PR with failing tests (actually just go ahead and do it, thanks!)

PR #968, opened 2024-10-15
fix(ux): improve image loading more smooth

konata33:
### Description

improve image loading more smooth
iShot Oct 16


Innei:
Great! There was still a problem here at the time. If it's a case of length > width, the ratio of the blurhash is not correct with the original image. Can you fix it together?

Innei:
#959

Innei:
@hyoban Can you track this PR?

PR #981, opened 2024-10-16
fix: Fix popup's width when viewport changes.

Przeblysk:
fix https://github.com/RSSNext/Follow/issues/958#issuecomment-2415633588 

The fix now is to use useViewport to get the component to re-render, and I'm not sure that's appropriate

### DIFF
--- a/apps/renderer/src/modules/entry-column/components/mark-all-button.tsx
+++ b/apps/renderer/src/modules/entry-column/components/mark-all-button.tsx
@@ -6,6 +6,7 @@ import { Trans, useTranslation } from "react-i18next"
 import { toast } from "sonner"
 import { useOnClickOutside } from "usehooks-ts"
 
+import { useViewport } from "~/atoms/hooks/viewport"
 import { ActionButton, Button, IconButton } from "~/components/ui/button"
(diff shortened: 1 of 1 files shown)


apps/renderer/src/modules/entry-column/components/mark-all-button.tsx
```diff
@@ -42,6 +43,10 @@ export const MarkAllReadWithOverlay = forwardRef<
   useOnClickOutside({ current: popoverRef }, () => {
     setShow(false)
   })
+
+  // change popup's width when viewport changes.
+  useViewport((v) => v.w)
```
Innei, line 48: Refactor `renderPopup` to a React Component, then move `useViewport` hook into the refactored component. 

The reason: this width change listener is only needed when Popup is shown. This is changed to improve performance.

Innei, line 48: Refactor `renderPopup` to a React Component, e.g `ConfirmPopup`, then move `useViewport` hook into the refactored component.

The reason: this width change listener is only needed when Popup is shown. This is changed to improve performance.

Innei (changes requested):

Innei (approved):

PR #983, opened 2024-10-16
feat: Inbox shows sender's address

yusixian:
### Description

Would like to be able to display the sender's address in the inbox, and the link to bounce from incoming emails can be a link to the sender, I checked the interface and found that there is this data

Before
image

After
image

Link to mailto:<email>
image

### Linked Issues

https://github.com/RSSNext/Follow/issues/805

### Additional context
apps/renderer/src/modules/entry-content/components/EntryTitle.tsx

```tsx
feed = useFeedById(entry?.feedId) as FeedModel
```
Why is this as a FeedModel? Testing revealed that it could be InboxModel


hyoban:
Does this affect the title of a normal entry?

yusixian:
> Does this affect the title of a normal entry?

No, I've tested the list and the normal feed streams

image

Triggered only if `feed?.type === “inbox”`.

PR #990, opened 2024-10-17
fix: trending drawer overflow

kovsu:
https://github.com/RSSNext/Follow/commit/3edee63a5a9f691481c8532ba524d541ca09c539#diff-2bd15723d33f49ade6aaf527ddfd0549ae37e38e6c5d456e4b88cb6da91c3415R142

This code seems lost. It makes the below problem.

image

### DIFF
--- a/apps/renderer/src/components/ui/scroll-area/ScrollArea.tsx
+++ b/apps/renderer/src/components/ui/scroll-area/ScrollArea.tsx
@@ -137,7 +137,10 @@ export const ScrollArea = React.forwardRef<
         <Viewport
           ref={setViewportRef}
           onWheel={stopPropagation}
-          className={cn(flex ? "[&>div]:!flex [&>div]:!flex-col" : "", viewportClassName)}
+          className={cn(
+            flex ? "[&>div]:!flex [&>div]:!flex-col" : "[&>div]:!block",
+            viewportClassName,
+          )}
           mask={mask}
         >
           {children}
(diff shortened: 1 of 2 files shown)


apps/renderer/src/components/ui/scroll-area/ScrollArea.tsx
```diff
@@ -137,7 +137,10 @@ export const ScrollArea = React.forwardRef<
         <Viewport
           ref={setViewportRef}
           onWheel={stopPropagation}
-          className={cn(flex ? "[&>div]:!flex [&>div]:!flex-col" : "", viewportClassName)}
+          className={cn(
+            flex ? "[&>div]:!flex [&>div]:!flex-col" : "[&>div]:!block",
```
Innei, line 141: You can't use `block`, it will cause an exception in the radix scrollbar.

Innei (changes requested):

apps/renderer/src/components/ui/scroll-area/ScrollArea.tsx
kovsu: What's the problem? Can you point out its details? 🥹

kovsu:
I think it would work fine here if I added this class to the trending drawer.

https://github.com/user-attachments/assets/69e6f301-ef22-4565-84aa-6d8c950c0ac7

Innei (approved):
It makes sense to modify only the class of this component, not the generic component.

PR #994, opened 2024-10-17
fix: optimize about share btn padding

winchesHe:
### Description

Before

https://github.com/user-attachments/assets/84908402-5ef3-464d-8e32-d66f7f2f59d0

After

https://github.com/user-attachments/assets/04258300-2ffa-4acd-9150-a604025b9c83


hyoban:
Thank you very much, but I think the current situation is okay

Innei:
@hyoban This fix is correct, it affects accessibility.

hyoban:
Sorry for my mistake, thanks for your contribution

winchesHe:
@Innei @hyoban 🫰

PR #1032, opened 2024-10-20
fix: not showing '0' when no followers on the search results card

GetToSet:
### Description

This PR fixes the issue that '0' is not shown when there's no followers on the search results card.

### Additional context

Screenshot 2024-10-20 at 21 37 42


lawvs (approved):
Nice catch

PR #1054, opened 2024-10-22
fix: margin between the top bar and the content

cuikaipeng:
### Description

This issue https://github.com/RSSNext/Follow/issues/1049 mentions the problem of the top bar being obscured. 
This PR https://github.com/RSSNext/Follow/pull/1051 addresses that issue.
But I noticed there are still some problems on the Video page.
The margins are too narrow. 
iShot_2024-10-22_10 50 22

Here's how it looked before.
iShot_2024-10-22_10 50 47

I tried to find the cause of the above issue. I checked the commit history, and in this commit https://github.com/RSSNext/Follow/commit/847c317bb4fe4258b7eed240e5e6422bf273838d#diff-8edd1371c19527656614803122cca80cb2fde924536821a4f868bbf727a55157, the "mb-2" style was removed, which led to the issue with the top bar being obscured. I think the removal of "mb-2" here,  is to handle the situation where there need a TimelineTabs component.

I think the approach to solving this issue should be to determine whether "mb-2" is needed based on the type of displayed content. 
So I submitted this PR.

### DIFF
--- a/apps/renderer/src/modules/entry-column/layouts/EntryListHeader.tsx
+++ b/apps/renderer/src/modules/entry-column/layouts/EntryListHeader.tsx
@@ -73,9 +73,12 @@ export const EntryListHeader: FC<{
   return (
     <div
       ref={containerRef}
-      className="flex w-full flex-col pl-6 pr-4 pt-2.5 transition-[padding] duration-300 ease-in-out"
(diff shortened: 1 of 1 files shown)


hyoban:
Wow, thank you so much for the PR.

PR #1081, opened 2024-10-23
feat: add horizontal scroll functionality to TimelineTabs

xhuajin:
### Description

**new feature**

TimelineTabs can scroll by mouse wheel without pressing shift. 

gif-demo: https://github.com/user-attachments/assets/e544d4e7-41b3-400c-a130-ad83614ef865

gif-demo

**fix**

Titles in timelinetabs cannot be displayed completely because of line-height:1.

img: https://github.com/user-attachments/assets/6dd6f1a2-dcf6-46af-81b4-8d9082436b04

image-demo


apps/renderer/src/modules/entry-column/layouts/TimelineTabs.tsx
```diff
@@ -20,9 +23,17 @@ export const TimelineTabs = () => {
 
   const navigate = useNavigateEntry()
   if (!hasData) return null
+
+  if (!listRef.current) return
+  const list = listRef.current
+  list.addEventListener("wheel", (e) => {
+    e.preventDefault()
+    list.scrollLeft += e.deltaY
+  })
```
hyoban, line 32: Why register event listeners during rendering?

hyoban (changes requested):

apps/renderer/src/modules/entry-column/layouts/TimelineTabs.tsx
xhuajin: 因为 timelineTabs 中的内容排列方向是 flex-direction: row，tab 如果太多的话，多余的部分会隐藏。想要看到后面的tab的内容需要按住 shift 再滚动。监听这里的滚动事件可以直接滚动查看溢出的 tab

Innei: You need to follow the React philosophy, writing it this way can lead to memory leaks

Innei (changes requested):

apps/renderer/src/modules/entry-column/layouts/TimelineTabs.tsx
xhuajin: Thanks for telling me, I used onWheel() to add the listener instead

hyoban:
Thanks for your contribution!

PR #1087, opened 2024-10-23
fix: Wrong initial position of tab animation

hamster1963:
### Description

https://github.com/user-attachments/assets/5887f22c-e0f4-4d0f-b465-7fbebfe71dae

Currently, under the power page, the tab indicator animation at the bottom will be shifted from its initial position after the page is scrolled, this can be avoided by setting style={{originY: ‘0px’,}}

### Additional context

This is solved by setting style={{originY: ‘0px’,}}

https://github.com/user-attachments/assets/1b95c099-54c5-4496-aa48-dda78804eb5b


hyoban:
Thanks for your contribution!

PR #1102, opened 2024-10-24
chore(i18n): add daily reward system translation

lc4t:
### Description

Add zh-CN translation for daily reward system


hyoban:
Thanks for your contribution

PR #1106, opened 2024-10-24
fix: EntryListHeader' margin in Pictures View

cuikaipeng:
### Description

Related: https://github.com/RSSNext/Follow/pull/1103
There are still some questions here
When there is no `TimelineTabs` component, there are some margin issues in the `Pictures Page`
iShot_2024-10-24_14 07 29

In my previous PR https://github.com/RSSNext/Follow/pull/1054, I added `mb-2 style` when there is not the `Articles Page` and `Pictures Page` because these two pages have `TimelineTabs` component. At that time, there are some issues with my modifications, I did not consider the situation where `TimelineTabs` component were not available (this component would not display when user do not have lists and inbox data)
So my idea is control whether `mb-2 style` is required by determining whether the `TimelineTabs` component is null
So I submitted this PR.


Innei:
`forwardRef` might not be a good choice, I just changed the CSS style

PR #1108, opened 2024-10-24
fix: optimize the appearing horizontal scroll bar.

hash-panda:
Daily summary feature, when long address information appears, an unfriendly horizontal scroll bar will appear. Optimize it.

image


Innei (approved):

Innei:
Thanks for contribution

PR #1109, opened 2024-10-24
feat: add system-ui font selection to desktop app

Aaron-212:
### Description

Add an option to use system-ui font on desktop app
image


Innei (approved):
Thanks for contribution

PR #1114, opened 2024-10-24
fix: solve overflow issue in achievement model

Shenhaoyi:
### Description
When the height is insufficient, the achievement list cannot be fully displayed.

before
image

after
image


hyoban:
How about applying the same scroll animation as the user profile, the upper part is a bit too high now

Innei (approved):

Innei:
I've modified it.

PR #1120, opened 2024-10-24
chore(i18n): update zh lang

ischanx:


Innei (approved):
Thanks for contribution

Issue #1127, opened 2024-10-25
能够更新，但每次更新会有下面的弹出提示。


gitwuneng:
### Describe the bug

[Window Title]
Error

[Main Instruction]
A JavaScript error occurred in the main process

[Content]
Uncaught Exception:
TypeError: Object has been destroyed
    at App.<anonymous> (file:///C:/Users/dell/AppData/Local/Follow/app-0.1.0-beta0/resources/app.asar/dist/main/index.js:96013:28)
    at App.emit (node:events:531:35)
    at App.<anonymous> (file:///C:/Users/dell/AppData/Local/Follow/app-0.1.0-beta0/resources/app.asar/dist/main/index.js:96050:13)
    at App.emit (node:events:531:35)
    at destroyMainWindow (file:///C:/Users/dell/AppData/Local/Follow/app-0.1.0-beta0/resources/app.asar/dist/main/index.js:43968:23)
    at quitAndInstall (file:///C:/Users/dell/AppData/Local/Follow/app-0.1.0-beta0/resources/app.asar/dist/main/index.js:56149:3)
    at Object.action (file:///C:/Users/dell/AppData/Local/Follow/app-0.1.0-beta0/resources/app.asar/dist/main/index.js:74645:5)
    at file:///C:/Users/dell/AppData/Local/Follow/app-0.1.0-beta0/resources/app.asar/dist/main/index.js:4710:20
    at WebContents.<anonymous> (node:electron/js2c/browser_init:2:86674)
    at WebContents.emit (node:events:519:28)

[确定]

### Feed Info

all

### Reproduction Video

111

### App Type

Electron / App

### Operating System

Windows 10

### Browser Info

Microsoft Edge 版本 130.0.2849.46 (正式版本) (64 位)

### Validations

- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.

### Contributions

- [ ] I am willing to submit a PR to fix this issue
- [ ] I am willing to submit a PR with failing tests (actually just go ahead and do it, thanks!)

Issue #1092, opened 2024-10-24
Update Error


hexWars:
### Describe the bug

点击更新后弹出如下弹窗

![image](https://github.com/user-attachments/assets/21b47a2b-1c9e-4de8-8983-86a792f433b7)

### Feed Info

无

### App Type

Electron / App

### Version

最新版

### Operating System

Windows 11

### Validations

- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.

### Contributions

- [ ] I am willing to submit a PR to fix this issue
- [ ] I am willing to submit a PR with failing tests (actually just go ahead and do it, thanks!)

harry10086:
一样，之前几个版本都是这样，而且只会删除老版本，不会安装新版本。
我都是重新安装一遍

hexWars:
> 一样，之前几个版本都是这样，而且只会删除老版本，不会安装新版本。 我都是重新安装一遍

我之前更新其实可以，只有这次不行

Innei:
Fixed in https://github.com/RSSNext/Follow/pull/1144

PR #1144, opened 2024-10-25
fix(app): throw error popup when close window

BlackHole1:
### Description

Once the window is destroyed, none of the window's methods can be used (except for `destroy` and `isDestroyed`).

### Linked Issues

Fix: #1127 #1092

### Additional context

The reason that destroy can be called multiple times is that Electron performs a check: https://github.com/electron/electron/blob/57920e7747f09627f99ed5c27457339aa7c3fe57/shell/common/gin_helper/destroyable.cc#L28

The ability to call isDestroyed is because no window methods are used during the implementation process: https://github.com/electron/electron/blob/57920e7747f09627f99ed5c27457339aa7c3fe57/shell/common/gin_helper/destroyable.cc#L48-L49

### DIFF
--- a/apps/main/src/index.ts
+++ b/apps/main/src/index.ts
@@ -123,7 +123,7 @@ function bootstrap() {
   app.on("before-quit", () => {
     // store window pos when before app quit
     const window = getMainWindow()
-    if (!window) return
+    if (!window || window.isDestroyed()) return
     const bounds = window.getBounds()
 
     store.set(windowStateStoreKey, {


BlackHole1:
This issue occurs intermittently, and the fundamental reason for this problem is that there are other places that are destroying the window. The code is as follows:

https://github.com/RSSNext/Follow/blob/8193cbcc51965af229761eeafaf37f6d969afb50/apps/main/src/updater/index.ts#L16-L19

https://github.com/RSSNext/Follow/blob/8193cbcc51965af229761eeafaf37f6d969afb50/apps/main/src/index.ts#L177-L180

Both pieces of code could trigger this issue, but the first one is more likely to do so. The second one is less frequent because it also listens to the `before-quit` event. In Electron, emitting `before-quit` is sequential (though it doesn't wait for the previous one to complete before executing the next, the code here is very simple, so it's almost never encountered, but it does exist objectively).

The first one triggers very frequently because it completely does not rely on any Electron events when destroying the window, leading to a scenario of "true concurrent execution."

hyoban:
Thank you so much for your help!!!

PR #1215, opened 2024-10-29
fix: disable `pointerDownOutside` trigger `onDismiss`

Debbl:
### Description

fix electron app
> I don't know why it can't be reproduced stably
- accidentally called `onClose` once more when opening the modal.
- `onClose` multiple called

https://github.com/user-attachments/assets/c3b9051f-86e7-4129-80ce-f57b67fea929

### DIFF
--- a/apps/renderer/src/components/ui/modal/stacked/modal.tsx
+++ b/apps/renderer/src/components/ui/modal/stacked/modal.tsx
@@ -94,15 +94,6 @@ export const ModalInternal = memo(
       onPropsClose?.(false)
     })
 
-    const onClose = useCallback(
-      (open: boolean): void => {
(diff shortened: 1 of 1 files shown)


Innei (changes requested):
This changes break press `Escape` to dismiss the modal

Debbl:
In the electron app, the `pointerDownOutside` may be triggered after calling `menu-closed`，so call `event.preventDefault` to disabled trigger `onDismiss`.
https://github.com/radix-ui/primitives/blob/74b182b401c8ca0fa5b66a5a9a47f507bb3d5adc/packages/react/dismissable-layer/src/DismissableLayer.tsx#L90

Innei (approved):

PR #1229, opened 2024-10-29
feat(integration): add outline integration

ericyzhu:
### Description

Added [Outline](https://www.getoutline.com/) integration. (Self-hosting instance is also supported)


Innei (approved):

PR #1269, opened 2024-10-31
fix: auto archived list flash

Innei:


hyoban:
https://github.com/user-attachments/assets/0262c972-c179-4185-9d0e-f45ac317d4ab

After I load the archived entry, I cannot load the timeline under the view.

PR #1348, opened 2024-11-03
feat(renderer): prevent currently executing async entry action from being executed again

ericyzhu:
### Description

Added a loading state indicator for the button of the currently executing async action, and prevented it from being clicked again during execution to avoid duplicate execution.

### PR Type

- [x] Feature
- [ ] Bugfix
- [ ] Hotfix

### Changelog

### DIFF
--- a/apps/renderer/src/modules/entry-content/header.tsx
+++ b/apps/renderer/src/modules/entry-content/header.tsx
@@ -57,20 +57,6 @@ function EntryHeaderImpl({
 
   const shouldShowMeta = (hideRecentReader || !isAtTop) && !!entryTitleMeta?.title
 
-  const [actionLoading, setActionLoading] = useState(
-    items.reduce((states, { key }) => {
-      states[key] = false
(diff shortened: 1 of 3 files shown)


apps/renderer/src/modules/entry-content/header.tsx
```diff
@@ -108,15 +122,25 @@ function EntryHeaderImpl({
...
+                  ) : item.icon ? (
                     <Slot className="size-4">{item.icon}</Slot>
                   ) : (
                     <i className={item.className} />
                   )
                 }
                 active={item.active}
                 shortcut={item.shortcut}
-                onClick={item.onClick}
+                onClick={async (e: React.MouseEvent<HTMLElement, MouseEvent>) => {
```
Innei, line 135: Instead of adding logic here, wrap the logic inside the ActionButton component.

Innei (changes requested):

PR #1384, opened 2024-11-04
fix: handle tipUsers migration for older feed versions

lawvs:
Fix https://github.com/RSSNext/Follow/issues/1308


apps/renderer/src/initialize/migrates/index.ts
```diff
@@ -9,14 +11,30 @@ declare global {
...
+  }
 
-  if (lastVersion && lastVersion !== APP_VERSION) {
-    appLog(`Upgrade from ${lastVersion} to ${APP_VERSION}`)
-    window.__app_is_upgraded__ = true
+  appLog(`Upgrade from ${lastVersion} to ${APP_VERSION}`)
+  window.__app_is_upgraded__ = true
 
+  try {
+    if (semver.lte(lastVersion, "0.1.2-beta.0")) {
```
Innei, line 33: No need to follow the version number, just execute according to the order of the migration program.

This is the first one, and there will be more in the future.

When the app starts, execute in order, record the completion time, and after executing once, it will not be executed again next time.

Innei (approved):

PR #1416, opened 2024-11-05
feat: add application tray functionality

lawvs:
Fix https://github.com/RSSNext/Follow/issues/177 https://github.com/RSSNext/Follow/issues/999

Introduce an application tray that allows users to interact with the main window through a system tray icon.


Innei:
macOS and Linux don't need tray and maybe only windows does

lawvs:
> macOS and Linux don't need tray and maybe only windows does

I will add an option that allows users to customize the tray behavior, such as minimizing to tray after closing the app.

Innei:
The menu icon on macOS is in grayscale?

lawvs:
> The menu icon on macOS is in grayscale?

Need to design new icons based on macOS [template images](https://developer.apple.com/documentation/appkit/nsimage/1520017-template).

PR #1422, opened 2024-11-05
fix: prevent withdrawal of zero amount to avoid unnecessary fees

abel533:
Added check in onSubmit to prevent withdrawal if amount is zero

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/RSSNext/Follow?shareId=789051f6-2228-4c4f-ac48-a593d371605d).

### DIFF
--- a/apps/renderer/src/modules/power/my-wallet-section/withdraw.tsx
+++ b/apps/renderer/src/modules/power/my-wallet-section/withdraw.tsx
@@ -69,6 +69,9 @@ const WithdrawModalContent = ({ dismiss }: { dismiss: () => void }) => {
   })
 
   const onSubmit = (values: z.infer<typeof formSchema>) => {
+    if (values.amount === 0) {
+      return
+    }
     mutation.mutate(values)
   }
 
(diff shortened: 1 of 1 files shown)


apps/renderer/src/modules/power/my-wallet-section/withdraw.tsx
```diff
@@ -69,6 +69,9 @@ const WithdrawModalContent = ({ dismiss }: { dismiss: () => void }) => {
   })
 
   const onSubmit = (values: z.infer<typeof formSchema>) => {
+    if (values.amount === 0) {
```
hyoban, line 72: We should perform data validation in form schema

hyoban (changes requested):

apps/renderer/src/modules/power/my-wallet-section/withdraw.tsx
abel533:
```js
  const formSchema = z.object({
    address: z.string().startsWith("0x").length(42),
    amount: z.number().min(0).max(cashablePowerTokenNumber),
  })
```
Modify min value directly? How much is appropriate?

hyoban: try `z.number().positive()`

hyoban:
Thanks for your contribution!

PR #1447, opened 2024-11-07
fix: prevent default scrolling behavior while using arrow keys to switch between entries

CanglongCl:
While using up and down arrow to switch between feed entries, the web will scroll a little. 

https://github.com/user-attachments/assets/a38dfba3-eeed-4213-bb63-32607247b33c

By preventing the default scrolling behavior of the arrow key, we avoid it. 

https://github.com/user-attachments/assets/2144e6f1-4a88-4018-89e3-61727b4923fe

### PR Type

- [ ] Feature
- [x] Bugfix
- [ ] Hotfix

### Changelog


Innei:

PR #1487, opened 2024-11-08
fix: update daily claim amount reference in tooltip

kovsu:
I wonder if it is better to remove the amount here. And actually `DAILY_CLAIM_AMOUNT.normal` isn't the real count I get today.

Before:

image

After:

image


hyoban:
I also think it would be better to remove it

Innei (approved):

Issue #1503, opened 2024-11-08
Auto-generated Default Title for New Feed, Easy to Edit


EnterMan123:
### Clear and concise description of the problem

When adding a new feed, the system will automatically generate a default title, and the user only needs to make necessary edits or adjustments in the title field. 

This is far more convenient than manually entering the title from the beginning.

👇example

https://github.com/user-attachments/assets/c29b2291-85f4-4123-a958-6b391cbe6ffb

### Suggested solution

X

### Validations

- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

EnterMan123:
https://github.com/user-attachments/assets/9bd42994-dcb3-40f4-953a-39c39c0973a1


如果无法添加这个，至少让标题变得可复制，就像 #1428

PR #1545, opened 2024-11-10
chore: default to using feed title to fill subscription modal input

ericyzhu:
### PR Type

- [ ] Feature
- [ ] Bugfix
- [ ] Hotfix
- [x] Other (please describe):

### Linked Issues

#1503

### DIFF
--- a/apps/renderer/src/modules/discover/feed-form.tsx
+++ b/apps/renderer/src/modules/discover/feed-form.tsx
@@ -202,8 +202,8 @@ const FeedInnerForm = ({
       form.setValue("view", `${subscription?.view}`)
       subscription?.category && form.setValue("category", subscription.category)
       form.setValue("isPrivate", subscription?.isPrivate || false)
-      form.setValue("title", subscription?.title || "")
     }
+    form.setValue("title", subscription?.title || feed.title || "")
   }, [subscription])
 
   const followMutation = useMutation({
@@ -290,7 +290,7 @@ const FeedInnerForm = ({
                   <FormDescription>{t("feed_form.title_description")}</FormDescription>
                 </div>
                 <FormControl>
-                  <Input {...field} />
+                  <Input {...field} placeholder={feed.title || ""} />
                 </FormControl>
                 <FormMessage />
               </FormItem>


apps/renderer/src/modules/discover/feed-form.tsx
```diff
@@ -202,8 +202,8 @@ const FeedInnerForm = ({
       form.setValue("view", `${subscription?.view}`)
       subscription?.category && form.setValue("category", subscription.category)
       form.setValue("isPrivate", subscription?.isPrivate || false)
-      form.setValue("title", subscription?.title || "")
     }
+    form.setValue("title", subscription?.title || feed.title || "")
```
Innei, line 206: I think this is unnecessary. cc @hyoban

ericyzhu: Just to fill the feed title into the input box by default without manual input for modification, if this change is not necessary, please close this pr.

ericyzhu: > I think this is unnecessary. cc @hyoban

Just to fill the feed title into the input box by default without manual input for modification, if this change is not necessary, please close this pr.

hyoban: This may cause the user's custom title to be updated to the feed title by default, instead of being null

MakiMixx83 (approved):

apps/renderer/src/modules/discover/feed-form.tsx
ericyzhu: > This may cause the user's custom title to be updated to the feed title by default, instead of being null

If the user input matches the feed title when submitting, change it to null.
What do you think?

We need a way to allow modification of part of the title without manual input to improve user experience.

hyoban: This is also not very good, what if the user wants to pin the custom title permanently? A button that can fill the default title would be better.

ericyzhu: > This is also not very good, what if the user wants to pin the custom title permanently? A button that can fill the default title would be better.

Okay, I think we need to consider how to design this ui interaction first.

PR #1624, opened 2024-11-16
fix: explicitly assign a value to `cancelId` for ‘Clear All Data' dialog

ericyzhu:
### Description

The Electron assigns the latin string "Yes" to the "Confirmation" action and "No" to the "Cancel" action. Although the display order of the buttons is reversed in the English UI, the value responded when pressing the ESC key is correct (index value of the "No" button). However, in the translated UI, the default value responded when pressing the ESC key is 0 (if the correct value is not assigned to `cancelId`), which leads to the data being cleared after pressing the ESC key.

Therefore, explicitly assign a value to `cancelId`.

image

### PR Type

- [ ] Feature
- [x] Bugfix
- [ ] Hotfix

### Changelog

### DIFF
--- a/apps/main/src/lib/cleaner.ts
+++ b/apps/main/src/lib/cleaner.ts
@@ -7,6 +7,7 @@ import { promisify } from "node:util"
 import { callWindowExpose } from "@follow/shared/bridge"
 import { app, dialog } from "electron"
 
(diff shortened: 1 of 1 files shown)


Innei (approved):

PR #1696, opened 2024-11-21
fix: add deps for integration cmds

lawvs:
Add dependencies to various command registration hooks to ensure proper functionality based on availability checks.

### DIFF
--- a/apps/renderer/src/modules/command/commands/integration.tsx
+++ b/apps/renderer/src/modules/command/commands/integration.tsx
@@ -93,6 +93,9 @@ const useRegisterEagleCommands = () => {
             }
           },
         }),
+    {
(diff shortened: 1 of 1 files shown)


Innei (approved):

apps/renderer/src/modules/command/commands/integration.tsx
```diff
@@ -93,6 +93,9 @@ const useRegisterEagleCommands = () => {
             }
           },
         }),
+    {
+      deps: [isEagleAvailable],
```
Innei, line 97: Let the `deps` parameter be like `useEffect`, passed directly as an array, and you can set related lint rules in ESLint.

PR #1708, opened 2024-11-22
fix: improve perform for feed column

lawvs:
Refactor components to enhance performance through memoization and improved state management.

Adjustments include changes to feed actions and modal handling, resulting in a more efficient rendering process.

### DIFF
--- a/apps/renderer/src/hooks/biz/useEntryActions.tsx
+++ b/apps/renderer/src/hooks/biz/useEntryActions.tsx
@@ -1,5 +1,5 @@
 import type { FeedViewType } from "@follow/constants"
-import { useCallback } from "react"
+import { useCallback, useMemo } from "react"
 
 import {
   getReadabilityStatus,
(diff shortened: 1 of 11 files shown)


apps/renderer/src/hooks/biz/useEntryActions.tsx
```diff
@@ -72,107 +72,120 @@ export const useEntryActions = ({ entryId, view }: { entryId: string; view?: Fee
   const isShowSourceContent = useShowSourceContent()
   const getCmd = useGetCommand()
   const runCmdFn = useRunCommandFn()
-  if (!entryId) return []
```
lawvs, line 75: Wrap with memo

apps/renderer/src/hooks/biz/useFeedActions.tsx
```diff
@@ -520,28 +518,3 @@ export const useRemoveFeedFromFeedList = (options?: {
     },
   })
 }
-
-export const useResetFeed = () => {
```
lawvs, line 524: Moved to the queries dir

apps/renderer/src/modules/feed-column/index.tsx
```diff
@@ -85,9 +85,9 @@ export function FeedColumn({ children, className }: PropsWithChildren<{ classNam
       setActive_(args)
 
       navigateBackHome(nextActive)
-      setSelectedFeedIds([])
+      setSelectedFeedIds(SELECT_NOTHING)
```
lawvs, line 88: use a stable ref

Innei (approved):

PR #1749, opened 2024-11-23
feat: add customizable columns for masonry view

### LINKED ISSUE #1243: Image/Video View, Adjust Grid Preview Size.

There is no way to do this at the moment, the image preview size is related to the program window size.

### Suggested solution

👇

https://github.com/user-attachments/assets/ba3477e5-9823-4d24-91af-936f091626e3

👇
- Control for the grid image size slider, similar to the effect found in Google Photos.
- Applicable to images and videos

### Alternative

_No response_

### Additional context

_No response_

### Validations

abel533:
### Description

Waterfall view supports adjusting the number of columns.

image

image

image

### PR Type

- [x] Feature
- [ ] Bugfix
- [ ] Hotfix

### Linked Issues

fix #1243 

### Changelog

### DIFF
--- a/apps/renderer/src/atoms/settings/ui.ts
+++ b/apps/renderer/src/atoms/settings/ui.ts
@@ -38,6 +38,7 @@ export const createDefaultSettings = (): UISettings => ({
 
   // View
   pictureViewMasonry: true,
+  pictureViewMasonryColumn: 0,
   pictureViewFilterNoImage: false,
   wideMode: false,
 })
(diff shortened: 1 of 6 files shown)


Innei:
Displaying the adjustment of column numbers directly on the toolbar does not seem very reasonable.

Perhaps using HoverCard would be better.

When hovering over the Masonry Button, a Floating Card is summoned.

Innei:
Thank you for your contribution, but I think this interaction is not good enough, so I have rewritten it. 0e0ce843235f01f33f4c5b9708aa67dac5901b46

PR #1760, opened 2024-11-23
fix(readability): remove <a> tags around <img> elements

### LINKED ISSUE #1757: When clicking an image, sometimes it will jump and open in browser.
### Describe the bug

我注意到，有时候在RSS阅读器里，使用Readability排版的文章中，点击文章里的图片时，本意是想点击查看图片，但总是会跳到浏览器中。

比如蓝点网的订阅源 https://www.landiannews.com/feed ，随便点进去一篇带图的文章，点击一张图片。

https://github.com/user-attachments/assets/4e8944b2-1562-49f2-8baf-a1dcef9f93c0

后来，我找到了这个bug的原因：在原文的 html 中，如果图片被 `<a>` 标签包裹，那么它就会出现如上bug。如果图片没有被 `<a>` 标签包裹，就不会出现这个bug。

Image

一种可能的修复方法：
Readability 中读取到的正文 html 可以再进一步处理，如果检测到`img`标签被`<a>`标签嵌套，就去掉外面的`<a>`标签，避免跳到浏览器中打开的bug，这个可以使用 JSDOM 或者 Rehype Plugin 实现

### Feed Info

https://www.landiannews.com/feed

### Reproduction Video

如上所示

### Environment

**App Version**: 0.2.3-beta.0
**OS**: Windows
**User Agent**: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Follow/0.2.3-beta.0 Chrome/130.0.6723.59 Electron/33.0.2 Safari/537.36
**Env**: electron
**Browser**: Chrome
**Session Trace Id**: `fQ66Q0pkbwZcAICC5O3KD`

PrinOrange:
Fixes #1757

### Description

Try to fix the issue #1757.

### PR Type

- [ ] Feature
- [x] Bugfix
- [ ] Hotfix

### Linked Issues
#1757

### Changelog


ericyzhu:
This will break the links in the original article.

PrinOrange:
In fact, most websites will add another layer of link to the image element in the article, pointing to the source of the image.
Like this

```html
<a src="https://example.com/image.png">
    <img src="https://example.com/image.png" />
</a>
```
The link's href is equals to the media source.

This will cause the bug in #1757.

As for what you said about breaking the article link, I made further adjustments: if the `img` element is wrapped by an `a` tag, and the `href` link of the `a` tag is equlas to the image link, then remove the outer `a` tag.

```html
<a src="https://example.com/image.png">
    <img src="https://example.com/image.png" />
</a>

<!-- Above code will remove the outer link tag, it will become to following code -->
<img src="https://example.com/image.png" />
```

If the image source link is different from the outer href link, the outer tag will not be removed.

```html
<a src="https://example.com">
    <img src="https://example.com/image.png" />
</a>
<!-- Above code will not remove the outer link tag. -->
```

hyoban:
Does this problem not only occur in readability? Maybe we also need to do similar processing on the content output by RSS.

PrinOrange:
> This issue also exists in the normal Entry Content, can it be handled together?

I have write a rehype plugin to remove wrapper links for image.
```ts
/**
 * rehype plugin to optionally unwrap <a> tags when their href matches the <img> src inside.
 */
function rehypeUnwrapImageLinks() {
  return (tree) => {
    visit(tree, "element", (node, index, parent) => {
      if (node.tagName === "a" && node.children && node.children.length === 1) {
        const child = node.children[0]
        // Check if the child is an <img> element
        if (
          child.tagName === "img" &&
          node.properties?.href &&
          child.properties?.src &&
          node.properties.href === child.properties.src
        ) {
          // Replace <a> with its <img> child
          parent.children.splice(index, 1, child)
        }
      }
    })
  }
}
```

And use them in whole `parseHtml` function.
```ts
  const pipeline = unified()
    // ....
    .use(rehypeUnwrapImageLinks)
    // ...
```

lawvs:
Thanks a bunch for putting in the effort on this PR! And I'm sorry, we decided to go with a different solution for this issue, but I appreciate your hard work. Thanks for understanding!

PR #1958, opened 2024-12-01
fix: handle undefined arguments in executeJavaScript call

lawvs:
Fix SyntaxError

```js
globalThis.__follow?.toast.error?.(,{"duration":1000})
//                                 ^ missed undefined
```

Also related to https://github.com/RSSNext/Follow/commit/7bdf10050cef7f39b0f485b89ca0445c49116d81 https://github.com/Migushthe2nd/MsEdgeTTS/issues/20


packages/shared/src/bridge.ts
```diff
@@ -89,7 +89,12 @@ function createProxy<T extends RenderGlobalContext>(window: BrowserWindow, path:
 
       try {
         return await window.webContents.executeJavaScript(
-          `(async () => { try { return await globalThis.${PREFIX}?.${methodPath}?.(${args.map((arg) => JSON.stringify(arg)).join(",")}) } catch (err) { console.error('Failed to executeJavaScript: ${methodPath}', err) } })()`,
```
lawvs, line 92: The try-catch clause cannot catch syntax errors.

Innei (approved):

PR #1972, opened 2024-12-02
feat(integration): Add readeck integration

syfxlin:
### Description

Added [Readeck](https://readeck.org/en/) integration.

Tips:
- Readeck supports self-hosting, so there is an additional endpoint in settings.
- Readeck does not support CORS by default, so you need to add CORS protocol support to the reverse proxy. The easiest way is to set the endpoint to the following format: `https://test.cors.workers.dev/?https://<your_readeck_domain>`.

### PR Type

- [x] Feature
- [ ] Bugfix
- [ ] Hotfix

### Additional context

- Missing some i18n support, I'm not familiar with other languages so I didn't add it.

### Changelog


lawvs (approved):
Nice works!

PR #1987, opened 2024-12-03
fix: optimize audio source comparison in audio player

xiangzy1:
### Description

When there is Chinese in the link, the audio src comparison is always false, causing the audio to play from the beginning

### PR Type

- [ ] Feature
- [x] Bugfix
- [ ] Hotfix

### Changelog

### DIFF
--- a/apps/renderer/src/atoms/player.ts
+++ b/apps/renderer/src/atoms/player.ts
@@ -81,7 +81,10 @@ export const AudioPlayer = {
       listId: routeParams.listId,
     })
 
-    if (this.audio.src !== v.src) {
+    const currentUrl = new URL(this.audio.src, window.location.href).toString()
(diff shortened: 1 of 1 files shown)


apps/renderer/src/atoms/player.ts
```diff
@@ -81,7 +81,10 @@ export const AudioPlayer = {
       listId: routeParams.listId,
     })
 
-    if (this.audio.src !== v.src) {
+    const currentUrl = new URL(this.audio.src, window.location.href).toString()
```
Innei, line 84: Is there a risk of throw an error here?

xiangzy1: fixed

Innei (approved):

PR #2031, opened 2024-12-06
hotfix: readability's bug of failed to load document.

PrinOrange:
### Description

I'm so sorry for the pull-request #2000 , it solves the XSS security problem in electron, but also introduces another bug: in the new version of electron, `DOMPurify` requires a Window environment. But I ignored this, so readability may fail to load when loading articles.

So I now emergently release a PR to fix this problem. This is a pr for fixing the bug of sanitizing that causes readability failed to fetch document content.

**I sincerely apologize.**

### PR Type

- [ ] Feature
- [x] Bugfix
- [x] Hotfix

### Changelog


PrinOrange:
The bug mentioned above affects the version `0.2.6.beta`. It may affect normal use.

image

Now I fixed this problem.

https://github.com/user-attachments/assets/953da5ac-8c8a-48d5-a546-92e987cbe71f

Innei (approved):

PR #2038, opened 2024-12-07
fix: handle popstate for navigation in mobile entry content

CanglongCl:
### Description

Resolved issue where back navigation to the pending view would not behave correctly after changing orientation in some device.

Before: after changing the orientation, the `BACK` button is navigate to previous feeds, which is expected to navigate to all feeds view. 

https://github.com/user-attachments/assets/92237fe1-706a-44bd-9425-4bf9b1f3ca7a

After: now the back navigation will be force pushed to previous feed list. 

https://github.com/user-attachments/assets/53278b62-bd29-4da9-9925-253d2ec8cf54

For some iPad users, using **Landscape** orientation to browser the list and preview on right, while using **Portrait** orientation for reading provides excellent experience. This PR will improve experience in this situation. 

### PR Type

- [ ] Feature
- [x] Bugfix
- [ ] Hotfix

### Changelog

### DIFF
--- a/apps/renderer/src/modules/entry-content/index.mobile.tsx
+++ b/apps/renderer/src/modules/entry-content/index.mobile.tsx
@@ -5,13 +5,13 @@ import { stopPropagation } from "@follow/utils/dom"
 import { cn } from "@follow/utils/utils"
 import { ErrorBoundary } from "@sentry/react"
 import { useEffect, useMemo, useState } from "react"
-import { useNavigate } from "react-router"
 
 import { useShowAITranslation } from "~/atoms/ai-translation"
 import { useAudioPlayerAtomSelector } from "~/atoms/player"
 import { useGeneralSettingSelector } from "~/atoms/settings/general"
 import { useUISettingKey } from "~/atoms/settings/ui"
 import { ShadowDOM } from "~/components/common/ShadowDOM"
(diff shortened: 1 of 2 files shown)


apps/renderer/src/modules/entry-content/index.mobile.tsx
```diff
@@ -40,6 +41,25 @@ export const EntryContent: Component<{
   compact?: boolean
   classNames?: EntryContentClassNames
 }> = ({ entryId, noMedia, compact, classNames }) => {
+  const navigateEntry = useNavigateEntry()
+  const { entryId: _, ...params } = useRouteParamsSelector((route) => route)
+
+  useEffect(() => {
```
Innei, line 47: You can handle the return logic in the BackButton instead of through listening to popState.

Innei (changes requested):

apps/renderer/src/modules/entry-content/index.mobile.tsx
CanglongCl: > You can handle the return logic in the BackButton instead of through listening to popState.

The gesture of sliding back from the edge on mobile devices uses the browser's go back logic, so I would say listening on `popstate` is better.

PR #2075, opened 2024-12-09
feat: login with password

hyoban:
### PR Type

- [x] Feature
- [ ] Bugfix
- [ ] Hotfix

### Changelog


apps/renderer/src/modules/profile/update-password-form.tsx
```diff
@@ -0,0 +1,229 @@
...
+      </form>
+    </Form>
+  )
+}
+
+export const UpdatePasswordForm = () => {
+  const { data: hasPassword, isLoading } = useHasPassword()
+
+  if (isLoading) {
+    return null
```
Innei, line 222: Add fallback and avoid CLS

apps/server/client/pages/(login)/login.tsx
```diff
@@ -111,25 +125,95 @@ function Login() {
...
+                        loginHandler(key)
+                      }}
+                    >
+                      <i className={cn("mr-2 text-xl", authProvidersConfig[key].iconClassName)} />{" "}
+                      {t("login.continueWith", { provider: provider.name })}
+                    </Button>
+                  ))}
+              {!!authProviders?.credential && (
+                <div className="w-[320px] space-y-2">
+                  {!isCredentialProvider && (
```
Innei, line 148: The login Input should be located on top of the Form, and then the third-party login using the round icon should be placed under the Input.

Innei (changes requested):

apps/renderer/src/modules/profile/update-password-form.tsx
hyoban: This is the last item, there should be no problem

DIYgod (approved):

PR #2346, opened 2024-12-27
feat: html render for native

hyoban:
### PR Type

- [ ] Feature
- [ ] Bugfix
- [ ] Hotfix

### Changelog


apps/mobile/src/components/common/Html.tsx
```diff
@@ -0,0 +1,14 @@
+"use dom"
+import "../../../../../packages/components/assets/colors-media.css"
```
Innei, line 2: use monorepo

PR #2572, opened 2025-01-15
feat(rsshub): add minimum value to input field

kovsu:
### PR Type

- [ ] Feature
- [ ] Bugfix
- [ ] Hotfix

### Changelog

### DIFF
--- a/apps/renderer/src/modules/rsshub/set-modal-content.tsx
+++ b/apps/renderer/src/modules/rsshub/set-modal-content.tsx
@@ -131,6 +131,7 @@ export function SetModalContent({
                           inputMode="numeric"
                           pattern="[0-9]*"
                           max={12}
+                          min={hasPurchase ? 0 : 1}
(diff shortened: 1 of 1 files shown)


apps/renderer/src/modules/rsshub/set-modal-content.tsx
```diff
@@ -122,7 +122,7 @@ export function SetModalContent({
                   <FormControl className="!mt-0">
                     <div className="flex items-center gap-10">
                       <div className="space-x-2">
-                        <Input className="w-24" type="number" max={12} {...field} />
+                        <Input className="w-24" type="number" max={12} min={1} {...field} />
```
DIYgod, line 125: In fact, if it is already purchased, the minimum value can be 0.

`hasPurchase ? 0 : 1`

DIYgod (changes requested):

PR #2576, opened 2025-01-15
fix: enhance shortcut handling for OS compatibility

kovsu:
### Description

https://discord.com/channels/1243823539426033696/1319455653005230191/1328696880154022010

### PR Type

- [ ] Feature
- [ ] Bugfix
- [ ] Hotfix

### Changelog

### DIFF
--- a/apps/renderer/src/constants/shortcuts.ts
+++ b/apps/renderer/src/constants/shortcuts.ts
@@ -117,26 +117,22 @@ const shortcutConfigs = {
       key: "Meta+K",
     },
   },
-} as const satisfies Shortcuts
+} as const
 
 function transformShortcuts<T extends Shortcuts>(configs: T) {
-  const result: Shortcuts = {}
+  const result = configs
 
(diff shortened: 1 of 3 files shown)


apps/renderer/src/constants/shortcuts.ts
```diff
@@ -116,6 +119,25 @@ export const shortcuts = {
...
+        name: config.name,
+        key: transformShortcut(config.key),
+      }
+    }
+  }
+
+  return result
+}
+
+export const shortcuts = transformShortcuts(shortcutConfigs)
```
Innei, line 139: This results in the loss of the original type. We need the type to be inferred with the key

Innei (changes requested):

apps/renderer/src/constants/shortcuts.ts
kovsu: CleanShot 2025-01-16 at 09 37 19@2x

PR #2637, opened 2025-01-22
feat(feed-form): set default title in input if empty

kovsu:
### PR Type

- [ ] Feature
- [ ] Bugfix
- [ ] Hotfix

### Changelog


hyoban:
This is not very good, because the feed's title may change, but after setting it up, it will always be user-defined.

kovsu:
@hyoban 👌 thx

Issue #2786, opened 2025-02-17
Can't set a proxy to null after using it


IzumiMasaharu:
### Describe the bug

After I've used the proxy, I can't set it to null again, and when I set the proxy to null, the proxy doesn't update to the state of not applying the proxy.

### Feed Info

feedID: muyunze2009@gmail.com

### Reproduction Video

<img width="1164" alt="Image" src="https://github.com/user-attachments/assets/f199373d-5fae-4288-bc2a-6fe28a8f15d2" />

### Environment

macOS Sequoia 15.3.1

### Validations

- [x] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.
- [x] This issue is valid

### Contributions

- [x] I am willing to submit a PR to fix this issue
- [ ] I am willing to submit a PR with failing tests (actually just go ahead and do it, thanks!)

IzumiMasaharu:
I posted this issue because I have the same problem as #2679 .
I think we can solve #2679 at the same time if the app allow us to change our proxy to null.

IzumiMasaharu:
The issue has been solved in the new version by merging my PR https://github.com/RSSNext/Follow/pull/2788.

PR #2788, opened 2025-02-17
fix: Resolved the issue of being unable to reset it to empty after using the proxy.

IzumiMasaharu:
### Description

Resolved the issue of being unable to reset it to empty after using the proxy.

### PR Type

- [ ] Feature
- [x] Bugfix
- [ ] Hotfix

### Linked Issues

#2786 

### Additional context

I would like to explain my method to fix the bug below.

- As I gradually delete the existing content in the Proxy input field in the Setting interface, the `proxy` option in the `~/Library/Application Support/Follow/db.json` file is updated in real time.
- However, when the content deleted from the Proxy field can no longer be parsed into a valid URL or becomes empty, the `proxy` in `db.json` stops updating.
- Similarly, if the content in the Proxy input field in the Setting interface is directly changed to a string that cannot be parsed as a valid URL, it will not trigger an update to `db.json`.

Therefore, it can be concluded that the function `setProxyConfig` handles both the empty string and invalid string inputs in the same way——not to update the `db.json` file.

To differentiate between these two cases, I chose to modify the return value of the function `normalizeProxyUri` called within function `setProxyConfig`. Specifically, when the input is an empty string, I made the function `normalizeProxyUri` return an empty string. This allows the function `setProxyConfig` to update the `proxy` field in `db.json` to an empty string via `store.set`.

**Note: Since I haven’t set up the related environment on my computer, I hope someone else can help with building it and testing whether my method can actually resolve Issue #2786 without causing any other issues.**

### Changelog


lawvs (approved):
Nice catch!

Innei (approved):

PR #2793, opened 2025-02-18
feat(select): add itemClassName prop for customizable item styling

kovsu:
### Description

CleanShot 2025-02-18 at 09 06 33@2x


apps/renderer/src/modules/settings/tabs/apperance.tsx
```diff
@@ -174,7 +175,13 @@ const ShikiTheme = () => {
           }
         }}
         triggerClassName="w-48"
-        renderItem={(item) => <span className="capitalize">{item.label}</span>}
+        renderItem={(item) =>
+          isMobile ? (
+            capitalizeFirstLetter(item.label)
```
kovsu, line 180: Only mobile has the error in console. Because it's native `select`

Innei (approved):

PR #2801, opened 2025-02-18
feat(mobile): audio player

hyoban:
### Description

resolves FOL-1469

### PR Type

- [ ] Feature
- [ ] Bugfix
- [ ] Hotfix

https://github.com/user-attachments/assets/5c5ada96-df63-4738-9df6-c3fc9598684f

### Changelog


apps/mobile/src/lib/player.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export async function pause() {
+  await TrackPlayer.pause()
+}
+
+export async function reset() {
+  await TrackPlayer.reset()
+}
+
+export async function seekBy(offset: number) {
```
Innei, line 53: Create a player instance, wrapping it with a class. Directly exporting functions doesn't seem to be very suitable here.

PR #2836, opened 2025-02-21
refactor(entries): remove isArchived parameter from entry fetching logic

hyoban:
### PR Type

- [ ] Feature
- [ ] Bugfix
- [ ] Hotfix

### Linked Issues

part of FOL-1626

- https://github.com/RSSNext/follow-server/pull/250

### Changelog

### DIFF
--- a/apps/renderer/src/store/entry/store.ts
+++ b/apps/renderer/src/store/entry/store.ts
@@ -113,7 +113,6 @@ class EntryActions {
     read,
     limit,
     pageParam,
-    isArchived,
   }: {
     feedId?: number | string
     inboxId?: number | string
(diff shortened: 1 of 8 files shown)


apps/renderer/src/store/entry/store.ts
```diff
@@ -370,7 +367,7 @@ class EntryActions {
             }
           }
 
-          if (item.settings && draft.flatMapEntries[item.entries.id] && !options?.isArchived) {
+          if (item.settings && draft.flatMapEntries[item.entries.id]) {
```
hyoban, line 370: We need to return the correct settings for the entries query in sever

PR #2948, opened 2025-03-03
feat(mobile): add deep link redirection to follow screen

lawvs:
Resolve FOL-1681

Implement deep link handling to redirect users to the follow screen when the app is opened with a specific URI scheme.

https://github.com/user-attachments/assets/8b30c04b-cf54-42a9-9e04-b5855b314182

### DIFF
--- a/apps/mobile/src/screens/+native-intent.tsx
+++ b/apps/mobile/src/screens/+native-intent.tsx
@@ -1,0 +1,23 @@
+// Redirects to the home screen when the app is opened with a deep link.
+// Test this by running the following command in the terminal:
+// pnpx uri-scheme open 'follow://add?id=1' --ios
+//
+// See https://docs.expo.dev/router/advanced/native-intent/#rewrite-incoming-native-deep-links
+
+export function redirectSystemPath(_options: { path: string; initial: boolean }) {
+  const { path } = _options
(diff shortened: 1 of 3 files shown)


apps/mobile/src/screens/+native-intent.tsx
```diff
@@ -0,0 +1,23 @@
...
+  const parsedUrl = parseUrl(path)
+  if (!parsedUrl) return
+  const id = parsedUrl.searchParams.get("id")
+  if (!id) return
+  return `/follow?id=${id}`
+}
+
+const parseUrl = (url: string) => {
+  try {
+    return new URL(url)
```
Innei, line 19: Is `URL.canParse` support?

lawvs: No, custom URL schemes like follow:// aren't recognized as valid by the standard URL API without some modifications. 

```ts
URL.canParse('follow://add?id=1')
// false
```

lawvs: No, custom URL schemes like follow:// aren't recognized as valid by the standard URL API without some modifications. 

```ts
URL.canParse('follow://add?id=1')
// false
```

PR #2953, opened 2025-03-04
feat(desktop): hide progress when reading percent is zero

kovsu:
Before:

After:


Innei:
I feel that there is a need for a placeholder, even though the current progress is 0.

kovsu:
> I feel that there is a need for a placeholder, even though the current progress is 0.

I think when it's zero. It's clean that doesn't show the progress. Anyway feel free to close this pr.

PR #3265, opened 2025-03-24
fix: improve skeleton loading states in feed preview

lawvs:
Resolve FOL-1719

Enhance the user interface by updating the follow button text and improving the EntryItemSkeleton component to dynamically handle the count of skeleton items. Additionally, add a skeleton loading state for entries in the ViewSelectorRadioGroup.


apps/desktop/src/renderer/src/modules/profile/user-profile-modal.shared.tsx
```diff
@@ -175,7 +175,7 @@ const SubscriptionItem: FC<{
               ) : (
                 <>
                   <FollowIcon className="mr-1 size-3" />
-                  {APP_NAME}
```
lawvs, line 178: Screenshot 2025-03-24 at 16 07 05

PR #3282, opened 2025-03-25
fix: adjust flex styles in NavigationHeader layout

lawvs:
Resolve FOL-1773

Screenshot 2025-03-26 at 21 22 19

Screenshot 2025-03-26 at 21 22 31

Refine the flex styles in the NavigationHeader component to improve layout consistency and responsiveness.


Innei (changes requested):
The title should be centered on the screen, if space permits.

PR #3284, opened 2025-03-25
refactor: enhance toolbar customization and mobile view actions

lawvs:
Streamline the order of toolbar actions using a Set, add a customization option for toolbar actions in settings, and ensure only the main action displays in mobile view.

image

### DIFF
--- a/apps/desktop/src/renderer/src/modules/entry-content/header.mobile.tsx
+++ b/apps/desktop/src/renderer/src/modules/entry-content/header.mobile.tsx
@@ -19,6 +19,7 @@ import { useEntry } from "~/store/entry/hooks"
 
 import { useCommand } from "../command/hooks/use-command"
 import type { FollowCommandId } from "../command/types"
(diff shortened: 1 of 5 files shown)


apps/desktop/src/renderer/src/modules/entry-content/header.mobile.tsx
```diff
@@ -26,16 +25,7 @@ import type { EntryHeaderProps } from "./header.shared"
...
-    () =>
-      [...sortedActionConfigs.mainAction, ...sortedActionConfigs.moreAction].filter(
-        (item) =>
-          !([COMMAND_ID.entry.copyLink, COMMAND_ID.settings.customizeToolbar] as string[]).includes(
-            item.id,
-          ),
-      ),
-    [sortedActionConfigs.mainAction, sortedActionConfigs.moreAction],
-  )
+  const actionConfigs = sortedActionConfigs.mainAction
```
lawvs, line 28: Only show main action in mobile view

PR #3385, opened 2025-04-01
feat: add Cubox integration for saving entries

### LINKED ISSUE #1070: Add cubox support in the integration

It is hoped that the configuration of Cubox will be integrated, so that articles can be quickly saved to Cubox in the reader.
Documents：https://help.cubox.pro/save/89d3/

### Suggested solution

1. Integrating with Cubox API
2. Supports custom URL scheme, users can configure to save to which APP.

### Alternative

_No response_

### Additional context

_No response_

### Validations

LionChenA:
close #1070 .
Added the feat on existing code, with i18n for English and Chinese.
- Add "Save to Cubox" feature
- Implement Cubox API integration
- Support Cubox API key configuration in settings
- Add localization for Cubox-related features

Questions:
- Cubox API requires Pro—would appreciate someone testing it.
- Not sure about the SVG—it’s AI-generated for now.

CleanShot 2025-04-01 at 20 21 00@2x


hyoban:
I see that Cubox has two APIs. Do you think adding an option to let users choose whether to save the web page or the content is necessary?

LionChenA:
> I see that Cubox has two APIs. Do you think adding an option to let users choose whether to save the web page or the content is necessary?

I've added calls for both APIs. When the memo mode is enabled, clicking the save button will check whether any text is selected — if so, it uses the memo API; otherwise, it falls back to the url API.

As for the SVG of Cubox, I’m currently in touch with their team. For now, it’s using the Eagle icon.

Lastly, here’s a preview.

CleanShot 2025-04-07 at 12 56 50

hyoban:
Thank you very much for what you did. Regarding the icon, I created a simplified version. If the cubox team provides an official icon, I will be happy to replace it, please feel free to create another PR to update it.

CleanShot 2025-04-08 at 17 21 09@2x

Issue #3379, opened 2025-03-31
Cannot click close button in the 'view source content' window


a7ca3:
### Platform

Desktop - Windows

### Describe the bug

When moving the mouse pointer to the close button, it becomes changing to a resize cursor and I cannot click it. This happens in the 'view source content' window.

### Reproduction Video

https://github.com/user-attachments/assets/3ea67b05-5086-4548-a4a2-6e258ca481df

### Validations

- [x] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.
- [x] This issue is valid

### Contributions

- [ ] I am willing to submit a PR to fix this issue
- [ ] I am willing to submit a PR with failing tests (actually just go ahead and do it, thanks!)

PR #3387, opened 2025-04-02
fix(desktop): modal close button cannot click

wilsonyiyi:
### PR Type

- [ ] Feature
- [x] Bugfix
- [ ] Hotfix

### Linked Issues
#3379 
### Changelog


kovsu:
Actually, it can be seen in the screen recording of the issue that dragging changes the size of the content inside rather than the dialog. Is this the default behavior officially?

wilsonyiyi:
> Actually, it can be seen in the screen recording of the issue that dragging changes the size of the content inside rather than the dialog. Is this the default behavior officially?

I also noticed this, it feels quite strange to operate, and I feel there is a problem. But this fix only addresses the issue of the close button not being clickable, as it also affects my own usage.

Innei (approved):

Issue #3388, opened 2025-04-02
条目链接获取错误


lazydao:
### Platform

Desktop - Windows

### Describe the bug

https://www.dongwm.com/atom.xml 这个订阅下的条目，获取链接时拿到的是 http://localhost:8000/post/*/ 

### Relevant Information

feed url：https://www.dongwm.com/atom.xml

### Environment

**App Version**: 0.4.0
**OS**: Windows
**User Agent**: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Folo/0.4.0 Chrome/132.0.6834.210 Electron/34.3.4 Safari/537.36
**Env**: electron
**Browser**: Chrome
**Session Trace Id**: `6iyZ4_3AX7deYORM2IPnL`
**Renderer**: 0.4.0

### Validations

- [x] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.
- [x] This issue is valid

### Contributions

- [ ] I am willing to submit a PR to fix this issue
- [ ] I am willing to submit a PR with failing tests (actually just go ahead and do it, thanks!)

DIYgod:
I believe the issue lies with the source RSS at specific times or under certain conditions, rather than with Folo. Furthermore, there is currently no deletion logic available for the RSS, so the problem cannot be resolved at this time.

PR #3389, opened 2025-04-02
feat(desktop): update EntryTitle to handle localhost url

kovsu:
Close: #3388


DIYgod:
Thank you for your contribution, but as mentioned in https://github.com/RSSNext/Folo/issues/3388#issuecomment-2772893314, I don't think this as an issue on Folo's side, and I don't want to add extra logic for this. 🫠

Innei (approved):

PR #3433, opened 2025-04-08
feat(ssr): enhance email validation and reCAPTCHA handling

0-o0:
### Description

There were several usability issues with the forget-password experience in app.follow.js that needed improvement:

1. **Email format validation enhancement**: Previously, validation feedback only appeared on mouse hover or after form submission when editing an email address. I've optimized this to validate the email format as users type, providing immediate feedback.

2. **reCAPTCHA logic and layout improvements**:
   - Changed the workflow logic: The previous implementation required users to click submit first before being challenged with reCAPTCHA. If users canceled the challenge, the submit button would remain in a perpetual loading state requiring page refresh.
   - Implemented a more standard approach where users must complete the reCAPTCHA challenge before the submit button becomes active, which prevents UI lockup issues.
   - Optimized the display of the rc-imageselect element to ensure it appears centered and fully visible on all screen sizes and configurations.

These changes create a more intuitive user experience with clearer feedback and prevent potential UI issues.

### PR Type

- [X] Feature
- [X] Bugfix
- [ ] Hotfix

### Changelog


apps/ssr/client/pages/(login)/forget-password.tsx
```diff
@@ -19,62 +19,134 @@ import { Input } from "@follow/components/ui/input/index.js"
...
-
 export function Component() {
   const { t } = useTranslation()
+  const navigate = useNavigate()
+  const recaptchaRef = useRef<ReCAPTCHA>(null)
+
+  const [captchaVerified, setCaptchaVerified] = useState(false)
+  const [captchaToken, setCaptchaToken] = useState<string | null>(null)
+  const [isExecutingCaptcha, setIsExecutingCaptcha] = useState(false)
+  const emailSchema = z
```
Innei, line 38: Extract to outside of this component

apps/ssr/client/pages/(login)/forget-password.tsx
```diff
@@ -104,26 +176,57 @@ export function Component() {
...
                   {t("login.submit")}
                 </Button>
               </div>
             </form>
           </Form>
         </CardContent>
       </Card>
+
+      <style
+        dangerouslySetInnerHTML={{
```
Innei, line 213: use `MemoedDangerousHTMLStyle` or write style in global.css. I think write in the global.css is better.

Innei (changes requested):

apps/ssr/client/pages/(login)/forget-password.tsx
0-o0: yeh,i have switched to using global css

DIYgod:
The use of invisible reCAPTCHA is to minimize the impact on users. For completely secure requests, reCAPTCHA will automatically pass in the background without the user noticing. According to our data, half of the requests will not trigger ReCAPTCHA, so I want to maintain the invisible status.

0-o0:
> The use of invisible reCAPTCHA is to minimize the impact on users. For completely secure requests, reCAPTCHA will automatically pass in the background without the user noticing. According to our data, half of the requests will not trigger ReCAPTCHA, so I want to maintain the invisible status.

I understand your question, but there is a little defect in user interaction when recaptcha is processed. For example, when I find an error in the mailbox and return to modify when there is a question, I will find that this submission has been in the loading state. Even if I just add a letter to the back of the mailbox, I can't submit it again, unless I refresh the web page and fill it in again. Now I have re-modified this part of the logic, kept invisible, and allow the user to return to the modification halfway to optimize the user experience.

apps/ssr/client/styles/index.css
```diff
@@ -30,3 +30,27 @@ button {
 [data-theme="dark"] {
   --bg-opacity: rgba(29, 29, 31, 0.72);
 }
+
+/* fix ReCAPTCHA layout */
+.g-recaptcha-bubble-arrow {
+  display: none !important;
+}
+
+div[style*="z-index: 2000000000"] {
```
DIYgod, line 39: The style writing here may not be very stable, can `src` be used to determine like this:

https://github.com/RSSNext/Follow/blob/0ccc57eee0e981a9f1cf52c4d139e07b53df9685/apps/desktop/src/renderer/src/styles/additional.css#L96

0-o0: i've applied your suggestion for the iframe's interaction styles,this uses the stable src selector specifically for pointer-events and -webkit-app-region，however, the centering styles (position: fixed, display: flex, etc.) must still target the container div[style*="z-index: 2000000000"]. css can't select parent elements, so i can't center the container by styling the child iframe,if don't need the layout change ,i can delete the added css

DIYgod: I haven't found a better solution either, but instead of simply not centering, I want to avoid introducing additional risks. Also, can the popup check be modified to detect iframe, or is it unnecessary altogether? https://github.com/RSSNext/Folo/pull/3433/files#diff-dd6c8a47c02d2c433be3f32ab5d29cce1d89c37d5182b08110ce10d41282619dR37

0-o0: I've removed the css code and refined the popup check and it should be fine now

DIYgod (approved):

DIYgod:
Thank you so much!

PR #3435, opened 2025-04-09
feat(desktop): update SearchCard layout

kovsu:
Related: https://github.com/RSSNext/Folo/commit/87d37f9f68a0bbb79479ea9a636ffbd7dd9d92b1

I think this new layout makes footer wide.

After:

image


Innei (changes requested):
Place preview button on the left of Follow button maybe is a great design.

kovsu:
What's about the search button. Should we unify the style?

PR #3458, opened 2025-04-10
feat(video-item): include entry content in modal preview

### LINKED ISSUE #3364: Need video description in Videos View

When I use the video view, it only contains the video title and cover, and I'm actually using follow to subscribe to the video content more to get updates than to watch it directly, so I'd like to support an option. When I open it, I can see the video's synopsis when I click on one of the videos in the video view. It will show for example, in a sidebar

### Suggested solution

I give the suggestion in previous part

### Alternative

_No response_

### Additional context

_No response_

### Validations

- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [x] This issue is valid

mandyschen:
### Description

This PR close #3364. I added a description underneath the videos (pulled from "content") when they are opened and are formatted to only include formatting from \<br> to protect against HTML injection.

### PR Type

- [X] Feature
- [ ] Bugfix
- [ ] Hotfix

Screenshot 2025-04-08 at 7 07 11 PM

https://github.com/user-attachments/assets/344df474-9d07-4ddd-9c27-1ca28b89bc1e

### Linked Issues

https://github.com/RSSNext/Folo/issues/3364 

### Changelog

- [X] I have updated the changelog/next.md with my changes.


hyoban:
The background of the text needs to be updated, the current one is unreadable

mandyschen:
> The background of the text needs to be updated, the current one is unreadable

Thanks for the feedback. Just added in a semi-transparent white background to make it more readable. 
image

Also fixed the issue with the any type.

PR #3461, opened 2025-04-11
refactor(desktop): improve shiki highlight component

kovsu:
| Before | After |
| ---- | ---- |
|  | |
|  |  |
| | |
|  |  |


Innei:
I think placing the language in the bottom right would be better. Try moving the top to the bottom? The background color of the copy button is a bit strange (in light mode).

kovsu:
> Try moving the top to the bottom?

🤔 Not good.

Innei:
Indeed, placing it at the top is better.

kovsu:
@Innei Update the copy-button variant. I think you can merge it now.

PR #3465, opened 2025-04-11
refactor(desktop):  move image gallery action to entry actions

kovsu:
Origin: https://discord.com/channels/1243823539426033696/1265925366820765818/1360146766665875478

Not sure if it conflicts with the previous design. Previously, the idea was to display this action only when the content had more than five images, so it can't be implemented in entry actions. But now that it's moved into entry actions, every item will have this action. It depends on how your team decides to design it.

https://github.com/user-attachments/assets/7d78689b-65a4-4d70-b23e-4182b9d66857


apps/desktop/src/renderer/src/modules/entry-content/actions/image-gallery-actions.tsx
```diff
@@ -1,35 +0,0 @@
...
-  const { t } = useTranslation()
-  const images = useEntry(id, (entry) => entry.entries.media)
-  const { present } = useModalStack()
-  if (images?.length && images.length > 5) {
-    return (
-      <ActionButton
-        onClick={() => {
-          tracker.entryContentHeaderImageGalleryClick({
-            feedId: id,
-          })
```
hyoban, line 21: We have tracker here

hyoban:
In addition, I think we can keep the previous logic and display only when there are enough images.

kovsu:
> In addition, I think we can keep the previous logic and display only when there are enough images.

I don't think so. If we move this to entry actions, it will show modal when we click this action whatever there are less or enough images.

hyoban:
At least hide this button when there is no image at all

PR #3558, opened 2025-04-23
fix(desktop): update isDirty prop correctly

kovsu:
If we add a new rule and then delete it, we actually don't have anything to save but the save button can be clicked.

https://github.com/user-attachments/assets/8f8ab51e-c8b7-480b-a702-9889fa58d93d


hyoban:
What if we delete a saved rule?

kovsu:
@hyoban  I have update the screen recording. Can you spend some time to review it?

hyoban:
I understand your idea, but I think the comparison here is unnecessary. There are not many cases where the modification is consistent with the original content, and there is also a performance overhead every time the rule is updated.

kovsu:
All right. 😢

PR #3613, opened 2025-04-28
feat(mobile): implement custom tab bar and item pressable components on android native

grtsinry43:
### Description

1.Adds native Android implementations for custom UI components previously available on iOS:

- **Custom Tab Bar:**
    - Implements `TabBarRootView` using `ViewPager2` and a custom `CustomTabBarView` (LinearLayout) to host RN-rendered tab items (`TabBarPortalView`).
    - Provides `TabScreenView` as a container for page content within Fragments managed by `TabBarPagerAdapter`.
    - Includes `TabBarModule`, `TabScreenModule`, and `TabBarPortalModule` for Expo autolinking and JS bridging (props: `selectedIndex`, event: `onTabIndexChange`, function: `switchTab`).
    - This approach mirrors the iOS version by avoiding the native `TabLayout` and allowing full customization of the tab bar via React Native components.

- **Item Pressable:**
    - Implements `ItemPressableView` extending `ExpoView`.
    * Provides native touch feedback using Android's foreground ripple (`selectableItemBackground`) by default.
    * Includes a prop `touchHighlight` to enable/disable feedback.
    * Fires `onItemPress` event on click/tap.
    * Includes `ItemPressableModule` for Expo integration.

2.Adjust content view padding to prevent overlap with Bottom Bar

### PR Type

- [x] Feature
- [x] Bugfix
- [ ] Hotfix

Screenshot_2025-04-28-15-37-15-306_is follow(1)

### Additional context

Since the project doesn't currently use these two native modules directly, it's difficult to see the results of changes right away. To maximize compatibility, I've aligned the Android implementation with the current iOS setup, matching parameters, events, and naming.

Separately, I've observed that the current Android beta feels a bit laggy or choppy during vertical scrolling and horizontal swipes. My next step is to delve into Android's native WebView rendering to try and boost the app's performance, integrating this work with functionalities like image preview.

### Changelog


Innei:
Perhaps you need to replace it in rn, then use Android's native components.

grtsinry43:
Okay,
Finally got some time to continue working on this PR 😂

I've replaced the "TabBar" parts with native components as planned.

It took quite a while to get the layout adjustments and lifecycle issues sorted out, mainly due to the significant differences between the Android layout system and how React Native handles things. However, those issues should be resolved now.

One important observation: using the Android Studio CPU Profiler, I noticed that `Choreographer.doFrame` consistently shows long execution times during TabScreen transitions (specifically the `flushMountItems` function seems to be a major contributor). This occurs in **both the original RN implementation and the current native components**, indicating that there's significant rendering overhead during these tab switches. Unfortunately, I'm not very experienced with React Native's animation libraries, so I wasn't able to dive deeper into optimizing that aspect.

Regarding the ItemPressable module, I wasn't entirely sure which existing module it was meant to replace. However, I substituted it during my testing, and didn't encounter any problems. You should be able to simply import and use this module directly as is.

**Best regards, and thanks to the Follow team for all your work!**

PR #3660, opened 2025-05-06
chore(desktop): minor improvement for discover page

kovsu:
Some minor improvements.

1. Add i18n

2. Fix eslint error

image

### DIFF
--- a/apps/desktop/layer/renderer/src/modules/discover/feed-card.tsx
+++ b/apps/desktop/layer/renderer/src/modules/discover/feed-card.tsx
@@ -92,7 +92,7 @@ export const FeedCard: FC<{
             )}
             <CardFooter className="flex justify-between gap-4 px-4 pb-3">
               <div className="flex items-center gap-3 text-sm text-zinc-500 dark:text-zinc-400">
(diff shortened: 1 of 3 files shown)


apps/desktop/layer/renderer/src/modules/discover/feed-card.tsx
```diff
@@ -141,6 +141,7 @@ export const FeedCard: FC<{
                 )}
                 <Button
                   variant={isSubscribed ? followedButtonVariant : followButtonVariant}
+                  disabled={isSubscribed}
```
DIYgod, line 144: It should not be disabled, the panel should be allowed to manage subscriptions.

DIYgod (changes requested):

apps/desktop/layer/renderer/src/modules/discover/feed-card.tsx
kovsu: 那这里 `已订阅` 改成 `编辑` 是不是更好。

DIYgod (approved):

Issue #3622, opened 2025-04-29
Special custom font won't work


Sisyphe42:
### Platform

Desktop - Windows, Desktop - Web

### Describe the bug

I tried to use "FzYouSongS 506L"(方正悠宋 简 506L) or "FzYouSongS 509R" as the UI font of Folo which all installed in the device, but they won't work, the actual font just stay as the former unchanged font. Havn't found other examples for the moment.

### Validations

- [x] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.
- [x] This issue is valid

### Contributions

- [ ] I am willing to submit a PR to fix this issue
- [ ] I am willing to submit a PR with failing tests (actually just go ahead and do it, thanks!)

kovsu:
I can't reproduce it.

https://github.com/user-attachments/assets/3bc091b8-d5d5-4b9b-a2bd-31219185167b

Sisyphe42:
> I can't reproduce it.
> 
>  CleanShot.2025-04-30.at.17.06.22.mp4

Only special fonts will reproduce it I think. I created a temp link where the font in my issue case can be downloaded: [https://tmp.link/f/6812295dbfe22](url)

PR #3674, opened 2025-05-07
fix(desktop): font names that include whitespace must be quoted

kovsu:
Close: #3622

Font family names that include whitespace or other special characters (like &, !, #) must be quoted. For example, "Times New Roman" or 'Arial Black'.

https://developer.mozilla.org/en-US/docs/Web/CSS/font-family#valid_family_names


Innei (approved):
Why not directly add a `"` to enclose it, this method should be universal

Issue #3693, opened 2025-05-09
Clicking the logo icon on the floating bar at the bottom of the mobile screen is invalid


lxmwwl:
### Platform

Mobile - Web

### Describe the bug

Clicking the logo in the floating bar does not pop up the RSS category main ban

### Validations

- [x] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.
- [x] This issue is valid

### Contributions

- [ ] I am willing to submit a PR to fix this issue
- [ ] I am willing to submit a PR with failing tests (actually just go ahead and do it, thanks!)

Innei:
Fixed in b031880b5941f722bd5decf347e9d8f47291b957

PR #3699, opened 2025-05-10
fix(ui): add id prop to ActionButton for improved accessibility

hqwuzhaoyi:
issues: https://github.com/RSSNext/Folo/issues/3693

### PR Type

- [ ] Feature
- [x] Bugfix
- [ ] Hotfix

CleanShot 2025-05-10 at 18 05 02@2x

FocuSee Project 2025-05-10 18-08-21

### Linked Issues
https://github.com/RSSNext/Folo/issues/3693
### Changelog

### DIFF
--- a/packages/internal/components/src/ui/button/action-button.tsx
+++ b/packages/internal/components/src/ui/button/action-button.tsx
@@ -78,6 +78,7 @@ export const ActionButton = ({
   const Trigger = (
     <button
       ref={buttonRef}
+      id={id}
       // @see https://github.com/radix-ui/primitives/issues/2248#issuecomment-2147056904
       onFocusCapture={stopPropagation}
       className={cn(


lawvs (approved):

lawvs (changes requested):
> The id [global attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes) defines an identifier (ID) which must be unique in the whole document.


Thank you for your contribution, but the current implementation does not guarantee this uniqueness within the broader context of the application.

> The purpose of the ID attribute is to identify a single element when linking (using a [fragment identifier](https://developer.mozilla.org/en-US/docs/Web/URI#fragment)), scripting, or styling (with [CSS](https://developer.mozilla.org/en-US/docs/Glossary/CSS)).

Furthermore, adding an `id` prop to the ActionButton component might not inherently improve accessibility.

PR #3722, opened 2025-05-13
feat(mobile): share intent support

hyoban:
### PR Type

- [ ] Feature
- [ ] Bugfix
- [ ] Hotfix

https://github.com/user-attachments/assets/f0de3752-0380-49c9-84d6-0a5177b9440b

### Changelog


Innei:
The effect of jumping to the app during the sharing process is not good, and it should use a floating ViewController to operate on the upper layer of the current app.

PR #3767, opened 2025-05-20
feat(desktop): unify with trending component

kovsu:
Before:

CleanShot 2025-05-20 at 16 52 21

After:

CleanShot 2025-05-20 at 16 52 49

CleanShot 2025-05-20 at 17 53 47

CleanShot 2025-05-20 at 17 55 18


apps/desktop/layer/renderer/src/modules/discover/recommendations.tsx
```diff
@@ -35,9 +35,9 @@ const LanguageOptions = [
     label: "中文",
```
Innei, line 35: Missing this?

Innei (changes requested):

apps/desktop/layer/renderer/src/modules/discover/recommendations.tsx
kovsu: Nope.

CleanShot 2025-05-20 at 17 37 02

kovsu: I think we can do it. As you can see, here is `English` not `英文` when language is `Chinese`
CleanShot 2025-05-20 at 17 46 09

Innei (approved):

PR #3870, opened 2025-06-05
fix(markdown): fix time was rendered as number and empty div

cscnk52:
### PR Type

- [ ] Feature
- [x] Bugfix
- [ ] Hotfix

Before:

image

After:

image

### Additional context

This issue is caused by `remark-directive` not rendering timestamps like `01:26` correctly — for example, it renders `01:26` as `01 <div></div>`.

Since I don't think the content rendered by this dependency is frequently used, I decided to remove it.

If `remark-directive` is indeed useful, feel free to close this PR — I’ll try to find another way to fix the issue.

Project homepage: [https://github.com/remarkjs/remark-directive](https://github.com/remarkjs/remark-directive)

### Changelog


cscnk52:
I made a patch to `micromark-extension-directive` to prevent strings like `:number` from being transformed into `<div></div>` elements.

Only one test fails in `micromark-extension-directive` repo, which is as expected:

image

cscnk52:
I have changed since last review, please re-review it.

PR #3893, opened 2025-06-09
feat(desktop): update unread count labels and descriptions for appear…

kovsu:


apps/desktop/layer/renderer/src/components/ui/modal/stacked/hooks.tsx
```diff
@@ -167,7 +167,7 @@ export const useDialog = (): DialogInstance => {
           ),
           content: ({ dismiss }) => (
             <div className="flex max-w-prose flex-col gap-3">
-              <div className="whitespace-pre">{options.message}</div>
+              <div className="whitespace-pre text-wrap">{options.message}</div>
```
kovsu, line 170: Before:

CleanShot 2025-06-09 at 11 46 07

After:

CleanShot 2025-06-09 at 11 45 46

apps/desktop/layer/renderer/src/modules/settings/helper/setting-builder.tsx
```diff
@@ -99,8 +99,8 @@ export const createSettingBuilder =
                     if (handled === "handled") {
                       return
                     }
-                    assertSetting.onChange(checked as T[keyof T])
                   }
+                  assertSetting.onChange(checked as T[keyof T])
```
kovsu, line 103: If we don't have `onChangeGuard` prop, the value won't be changed.

PR #4027, opened 2025-06-29
fix: fix social media click Read more refresh

cscnk52:
### Description

This PR contain two changes:

- because the 'read more' button in social-media is contained by a `<a>` tag, maybe this will cause a refresh? (i can't make sure, because dev version not refresh, but release version refresh, require further discussion, this is related to #4006 ), and change `SocialMediaItemSkeleton` to function component.

- add opacity for preview-media message, see picture:

image

### PR Type

- [ ] Feature
- [x] Bugfix
- [ ] Hotfix

### Changelog

### DIFF
--- a/apps/desktop/layer/renderer/src/modules/entry-column/Items/social-media-item.tsx
+++ b/apps/desktop/layer/renderer/src/modules/entry-column/Items/social-media-item.tsx
@@ -217,7 +217,7 @@ export function SocialMediaItemStateLess({ entry, feed }: EntryItemStatelessProp
   )
 }
 
-export const SocialMediaItemSkeleton = () => (
+export const SocialMediaItemSkeleton = (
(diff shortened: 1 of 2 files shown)


apps/desktop/layer/renderer/src/modules/entry-column/Items/social-media-item.tsx
```diff
@@ -217,7 +217,7 @@ export function SocialMediaItemStateLess({ entry, feed }: EntryItemStatelessProp
   )
 }
 
-export const SocialMediaItemSkeleton = (
+export const SocialMediaItemSkeleton = () => (
```
Innei: Why change this to `FC`

cscnk52: Because Vite warning for this component can't fast refresh, I will revert it.

Issue #4019, opened 2025-06-28
Bug: Clicking on links in social media does not open


lxmwwl:
### Clear and concise description of the problem

Unable to open the link

### Suggested solution

Hope it can be fixed

### Validations

- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [x] This issue is valid

DIYgod:
Thank you for the feedback. Could you provide more specific information? Such as which specific entry has an issue or provide a screenshot.

letthere-belight:
> Thank you for the feedback. Could you provide more specific information? Such as which specific entry has an issue or provide a screenshot.感谢您的反馈。您能提供更具体的信息吗？比如具体哪个条目有问题，或者提供一下截图。

As shown in the video, links in social media feeds cannot be opened. (Also, text in social media feeds cannot be selected and copied.)

https://github.com/user-attachments/assets/614fbe86-4f54-44bd-aa24-58193b578c8b

PR #4057, opened 2025-07-02
feat(desktop): enhance link handling

kovsu:
Close: #4019


Innei:
It is not an effective solution. Moreover, it leads to other problems. I will revert the code.

PR #4064, opened 2025-07-02
fix: enhance auth client with mobile-specific headers

lawvs:
Fix weird user agent on Android For Auth client.

Add mobile-specific headers to the auth client to improve compatibility and functionality on mobile devices.

This includes setting the User-Agent and other relevant headers based on the mobile environment.

### DIFF
--- a/apps/mobile/src/lib/auth.ts
+++ b/apps/mobile/src/lib/auth.ts
@@ -2,11 +2,16 @@ import { expoClient } from "@better-auth/expo/client"
 import { baseAuthPlugins } from "@follow/shared/auth"
 import { isNewUserQueryKey } from "@follow/store/user/constants"
 import { whoamiQueryKey } from "@follow/store/user/hooks"
+import { createMobileAPIHeaders } from "@follow/utils/headers"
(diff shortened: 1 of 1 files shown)


apps/mobile/src/lib/auth.ts
```diff
@@ -39,8 +44,30 @@ const plugins = [
...
+          OS: Platform.OS,
+          isPad: Platform.OS === "ios" && Platform.isPad,
+        },
+        installerPackageName: await DeviceInfo.getInstallerPackageName(),
+      })
+
+      Object.entries(headers).forEach(([key, value]) => {
+        ctx.headers.set(key, value)
+      })
+      ctx.headers.set("User-Agent", await getUserAgent())
```
lawvs, line 62: Android phones no longer send a user-agent like okhttp/xxx

Example user agent:

```
Dalvik/2.1.0 (Linux; U; Android 15; Pixel 9a Build/BP1A.2505xx.xxx.xx) Folo/0.2.2.1
```

Innei (approved):

PR #4066, opened 2025-07-03
feat(desktop): add missing i18n key

kovsu:


kovsu:
[Here](https://github.com/RSSNext/Folo/blob/dev/apps/desktop/layer/renderer/src/modules/profile/user-profile-modal/UserProfileModalContent.tsx#L280) has other problem. 



You can see `socialLinks` is `{}`, so it will make `UserSocialSection` visible though there is not any data.

```ts
if (!user || (!user.bio && !user.website && !user.socialLinks)) return null
```

But it only exists on when remote data is loading.

PR #4087, opened 2025-07-05
fix: download img support fallback src

cloudflypeng:
### PR Type

- [ ] Feature
- [x] Bugfix
- [ ] Hotfix

before
Kapture 2025-07-05 at 15 36 28
after
Kapture 2025-07-05 at 15 43 14

### Additional context

When preview image, the download button use the cors href, but the FallbackableImage has the currentSrc to Proxy image,
so I rember the href and open it when download.
让下载按钮也可以使用到代理之后的连接, 防止图片无法访问

### Changelog

### DIFF
--- a/apps/desktop/layer/renderer/src/components/ui/media/PreviewMediaContent.tsx
+++ b/apps/desktop/layer/renderer/src/components/ui/media/PreviewMediaContent.tsx
@@ -302,6 +302,14 @@ export const PreviewMediaContent: FC<{
   // This only to delay show
   const [currentSlideIndex, setCurrentSlideIndex] = useState(initialIndex)
 
+  // fallbackMap use in download to fix the cors image
+  const [fallbackMap, setFallbackMap] = useState<Record<string, string>>({})
+  const handleFallback = (original: string, fallback: string) => {
+    if (original && fallback) {
+      setFallbackMap((prev) => ({ ...prev, [original]: fallback }))
+    }
+  }
+
   useEffect(() => {
     if (emblaApi) {
       emblaApi.on("select", () => {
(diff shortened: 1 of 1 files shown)


apps/desktop/layer/renderer/src/components/ui/media/PreviewMediaContent.tsx
```diff
@@ -302,6 +302,14 @@ export const PreviewMediaContent: FC<{
   // This only to delay show
   const [currentSlideIndex, setCurrentSlideIndex] = useState(initialIndex)
 
+  // fallbackMap use in download to fix the cors image
+  const [fallbackMap, setFallbackMap] = useState<Record<string, string>>({})
```
Innei, line 306: Why is it a record?

apps/desktop/layer/renderer/src/components/ui/media/PreviewMediaContent.tsx
```diff
@@ -442,6 +457,12 @@ const FallbackableImage: FC<
     currentSrc === src ? "origin" : "proxy",
   )
 
+  useEffect(() => {
```
Innei, line 460: This `useEffect` may cause some rendering issues; it should not be used here.

Innei (changes requested):

apps/desktop/layer/renderer/src/components/ui/media/PreviewMediaContent.tsx
cloudflypeng: 我想记录原本url和回退的url的映射,但是我现在觉得下载功能是不是应该放在FallbackableImage 那个组件里,而不是在wrapper那个组件里,这样这个pr就可以关掉了,但是要修改一下FallbackableImage

PR #4103, opened 2025-07-08
Fix: GitHub Actions Workflow Vulnerable to Code Injection Attacks in .github/workflows/build-ios.yml

kira-offgrid:
**Context and Purpose:**

This PR automatically remediates a security vulnerability:
- **Description:** Using variable interpolation `${{...}}` with `github` context data in a `run:` step could allow an attacker to inject their own code into the runner. This would allow them to steal secrets and code. `github` context data can have arbitrary user input and should be treated as untrusted. Instead, use an intermediate environment variable with `env:` to store the data and use the environment variable in the `run:` script. Be sure to use double-quotes the environment variable, like this: "$ENVVAR".
- **Rule ID:** yaml.github-actions.security.run-shell-injection.run-shell-injection
- **Severity:** HIGH
- **File:** .github/workflows/build-ios.yml
- **Lines Affected:** 114 - 114

This change is necessary to protect the application from potential security risks associated with this vulnerability.

**Solution Implemented:**

The automated remediation process has applied the necessary changes to the affected code in `.github/workflows/build-ios.yml` to resolve the identified issue.

Please review the changes to ensure they are correct and integrate as expected.

### DIFF
--- a/.github/workflows/build-ios.yml
+++ b/.github/workflows/build-ios.yml
@@ -18,7 +18,7 @@ on:
         description: "Build profile"
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.event.inputs.profile }}
+  group: ${{ github.workflow }}-"$GITHUB_REF_VAR"-${{ github.event.inputs.profile }}
   cancel-in-progress: true
 
 jobs:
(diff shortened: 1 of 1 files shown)


hyoban:
Thank you for your contribution, but the changes in this PR have formatting issues, so the commit has been reverted

lawvs:
Could you provide a PoC regarding this issue?

PR #4124, opened 2025-07-10
refactor!: drop pwa

hyoban:
### PR Type

- [ ] Feature
- [ ] Bugfix
- [ ] Hotfix

### Changelog


Innei:
After deleting PWA, it is still necessary to retain a manifest.json. It greatly helps with the bookmark style added to the home screen.

Innei:
A very important issue needs to be considered. For users who have registered for service worker, the new version of pwa no longer exists. Can they update to the new version?

PR #4168, opened 2025-07-14
feat(desktop): return to back path in preview mode

kovsu:
https://github.com/user-attachments/assets/34861a1a-a058-461d-9a32-6750a5c3289c


apps/desktop/layer/renderer/src/modules/subscription-column/subscription-list/EmptyFeedList.tsx
```diff
@@ -35,7 +35,9 @@ export const EmptyFeedList = memo(({ onClick }: { onClick?: (e: React.MouseEvent
           onClick={handleClick}
         >
           <i className="i-mgc-arrow-right-up-cute-re text-3xl" />
-          <span className="text-balance text-sm">{t("sidebar.already_on_discover_page")}</span>
+          <span className="text-balance text-center text-sm">
+            {t("sidebar.already_on_discover_page")}
+          </span>
```
kovsu, line 40: Make text center

Image

apps/desktop/layer/renderer/src/modules/subscription-column/FeedItem.tsx
```diff
@@ -222,8 +222,9 @@ const FeedItemImpl = ({ view, feedId, className, isPreview }: FeedItemProps) =>
         >
           <i className="i-mgc-add-cute-re text-accent text-base" />
         </Button>
+      ) : (
+        <UnreadNumber unread={feedUnread} className="ml-2" />
```
kovsu, line 226: Avoid this situation. 
Title --- PlusIcon ---- UnreadNum

Image

kovsu: This is also the reason why we have to navigate to back path

Innei (approved):

PR #4188, opened 2025-07-16
refactor: simplify audio state handling

lawvs:
Enhance audio playback control by simplifying state management and enabling audio pause. Remove unnecessary properties and streamline the handling of audio states across components.

### DIFF
--- a/apps/mobile/src/components/ui/video/PlayerAction.tsx
+++ b/apps/mobile/src/components/ui/video/PlayerAction.tsx
@@ -34,6 +34,18 @@ export function PlayerAction({
     onPress()
   }, [onPress])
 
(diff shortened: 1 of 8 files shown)


apps/mobile/src/components/ui/video/PlayerAction.tsx
```diff
@@ -41,13 +45,13 @@ export function PlayerAction({
           intensity={30}
           experimentalBlurMethod="none"
         />
-        {isPlaying ? (
-          <PauseCuteFiIcon color="white" width={iconSize} height={iconSize} />
-        ) : isLoading ? (
-          <PlatformActivityIndicator />
-        ) : (
+        {mediaState === "paused" && (
```
Innei: switch...case maybe better

PR #4204, opened 2025-07-18
fix: button is not disabled when in starred list

yansq:
### Description

When in the starred list, buttons such as "Switch to Wide Mode," "Refresh," and "Show All" are hidden but remain clickable.

image

The CSS property `pointer-events: none;` does not affect child elements.

### PR Type

- [ ] Feature
- [x] Bugfix
- [ ] Hotfix

https://github.com/user-attachments/assets/2b6656c9-b7c0-4596-94c4-e37d2e1f2c4a

### Changelog


packages/internal/components/src/ui/button/action-button.tsx
```diff
@@ -107,7 +107,7 @@ export const ActionButton = ({
       // @see https://github.com/radix-ui/primitives/issues/2248#issuecomment-2147056904
       onFocusCapture={stopPropagation}
       className={cn(
-        "no-drag-region pointer-events-auto inline-flex items-center justify-center",
```
Innei: We need this css property.

Innei (changes requested):

packages/internal/components/src/ui/button/action-button.tsx
yansq: Maybe we should pass the className to the ActionButton?

apps/desktop/layer/renderer/src/modules/entry-column/layouts/EntryListHeader.tsx
```diff
@@ -83,7 +85,7 @@ export const EntryListHeader: FC<{
           <div
             className={cn(
               "text-text-secondary relative z-[1] flex items-center gap-1 self-baseline",
-              (isInCollectionList || !headerTitle) && "pointer-events-none opacity-0",
+              showButtons && "pointer-events-none opacity-0",
```
Innei:
```suggestion
              showButtons && "[&_*]:!pointer-events-none opacity-0",
```

apps/desktop/layer/renderer/src/modules/entry-column/layouts/EntryListHeader.tsx
```diff
@@ -97,9 +99,15 @@ export const EntryListHeader: FC<{
             )}
 
             <AppendTaildingDivider>
-              {!views[view]!.wideMode && <WideModeButton />}
-              {view === FeedViewType.SocialMedia && <DailyReportButton />}
-              {view === FeedViewType.Pictures && <SwitchToMasonryButton />}
+              {!views[view]!.wideMode && (
+                <WideModeButton className={showButtons ? "pointer-events-none" : void 0} />
```
Innei: No needed

Innei (changes requested):

Innei (approved):

PR #4284, opened 2025-07-31
feat(desktop): optimize PeekModal loading experience

kovsu:
Before: The modal shows up only after the content loads, with no loading state shown in between — that’s not a good ux.

After: Displaying a skeleton while loading improves the user experience.

<table>

<tr>

<td>

Before

<td>

After

<tr>

<td>

https://github.com/user-attachments/assets/c4a6e5c0-7279-4f35-8be9-d86f0ee37b40

<td>

https://github.com/user-attachments/assets/5ff2c5c8-ce79-4851-8a52-2902532a55cb

</table>


Innei (approved):
Wow!

PR #4488, opened 2025-09-16
chore: update dependencies

lawvs:
Update various package dependencies to their latest versions and remove the immer patch that is no longer needed.


apps/desktop/layer/main/package.json
```diff
@@ -28,22 +28,22 @@
     "@follow/shared": "workspace:*",
     "@follow/utils": "workspace:*",
     "@openpanel/web": "1.0.1",
-    "@sentry/electron": "6.8.0",
+    "@sentry/electron": "7.0.0",
```
lawvs: > MIGRATION.md 
> ## Upgrading from 6.x to 7.x
> Most users won't notice many breaking API changes in v7 but it's worth checking out the [JavaScript v10 migration guide](https://docs.sentry.io/platforms/javascript/migration/v9-to-v10/).

lawvs: Reverted, it breaks `@sentry__electron.patch`

Issue #4504, opened 2025-09-20
opening a photo in the message also plays the video


edwardchenchen:
### Platform

Desktop - Windows

### Describe the bug

this message is a photo reply to a video
https://x.com/manateelazycat/status/1969415379941540101
when imported into folo it has both the image and video, which is fine 
but
opening the photo in the message also plays the video

 

### Relevant Information

feed
rsshub://twitter/user/manateelazycat
message
https://x.com/manateelazycat/status/1969415379941540101

### Reproduction Video

https://github.com/user-attachments/assets/1c782f08-6d37-4b37-bd23-006f23267fb4

### Validations

- [x] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A, please open a GitHub Discussion instead.
- [x] This issue is valid

### Contributions

- [ ] I am willing to submit a PR to fix this issue
- [ ] I am willing to submit a PR with failing tests (actually just go ahead and do it, thanks!)

edwardchenchen:
this problem also exist on the website web view

PR #4531, opened 2025-09-30
feat: add video playback control in preview media

kovsu:
### Description

fix: #4504 

The `VideoPlayer` has `autoplay` attribute. So if we open the video in media, all videos will be playing.

### PR Type

- [ ] Feature
- [ ] Bugfix
- [ ] Hotfix

### Changelog

### DIFF
--- a/apps/desktop/layer/renderer/src/components/ui/media/PreviewMediaContent.tsx
+++ b/apps/desktop/layer/renderer/src/components/ui/media/PreviewMediaContent.tsx
@@ -21,6 +21,7 @@ import { ipcServices } from "~/lib/client"
 import { replaceImgUrlIfNeed } from "~/lib/img-proxy"
 
 import { useCurrentModal } from "../modal/stacked/hooks"
+import type { VideoPlayerRef } from "./VideoPlayer"
 import { VideoPlayer } from "./VideoPlayer"
 
 // Calculate the dynamic scale value and offset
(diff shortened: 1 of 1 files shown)


apps/desktop/layer/renderer/src/components/ui/media/PreviewMediaContent.tsx
```diff
@@ -342,8 +356,12 @@ export const PreviewMediaContent: FC<{
               <div className="mr-2 flex w-full flex-none items-center justify-center" key={med.url}>
                 {med.type === "video" ? (
                   <VideoPlayer
+                    ref={(el) => {
+                      if (videoRefs.current && el) {
+                        videoRefs.current.push(el)
```
Innei: missing cleanup func

Innei (changes requested):

apps/desktop/layer/renderer/src/components/ui/media/PreviewMediaContent.tsx
Innei: and pass memo func here

kovsu: Is this that you wanna see? Actually I'm not familiar with react, if is something wrong you can modify this PR directly.

CleanShot 2025-10-01 at 13 51 46@2x

PR #4622, opened 2025-10-27
refactor: enhance session management and in AI chat

lawvs:
Improve session synchronization logic, delete remote chat sessions when clearing empty sessions, and enhance task session retrieval with better error handling during test runs. Adjust chat ID matching for task sessions and improve unread message detection.


lawvs:
In favor of https://github.com/RSSNext/Folo/pull/4624

PR #4670, opened 2025-11-07
fix: handle chatId encoding in reconnectToStream

lawvs:


apps/desktop/layer/renderer/src/modules/ai-chat/store/transport.ts
```diff
@@ -132,4 +132,13 @@ class ExtendChatTransport extends HttpChatTransport<BizUIMessage> {
       console.error("Failed to persist generated title:", error)
     }
   }
+
+  override reconnectToStream(
+    options: Parameters<HttpChatTransport<BizUIMessage>["reconnectToStream"]>[0],
+  ) {
+    if (options.chatId.includes("/")) {
```
Innei: Not only `/` but also `%` `&` etc.

PR #4926, opened 2026-03-21
feat(mobile): add horizontal swipe navigation between articles

### LINKED ISSUE #4577: Mobile UX:  Next Article Swipe Left

Currently in read article mode, you can only swipe up to proceed to the next article if you are at the very bottom of the article.   I think it would be a better UI experience to be able to swipe left or right to proceed to the next/previous article.

Some articles are very lengthy which makes the swipe up function not practical in those cases.  Additionally the left/right swipe experience would follow other feed apps such as inoreader and feedly.

### Suggested solution

Change the swipe up (available only upon reaching the very bottom of the article), to swipe left/right to proceed to the next/previous article.

### Alternative

_No response_

### Additional context

_No response_

### Validations

- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.
- [x] This issue is valid

atmasphere:
## Summary

- Adds swipe left/right to navigate between articles in the entry detail screen
- Adds a subtle prev/next navigation bar with position counter at the bottom of articles
- Extracts shared `navigateToEntry()` function used by swipe, tap buttons, and existing pull-up-to-next

## Problem

Currently in article reading mode, the only way to advance to the next article is to scroll all the way to the bottom and pull up. This is impractical for long articles and doesn't match the interaction patterns of other RSS readers (Reeder, Feedbin, Inoreader).

Closes #4577

### New files
- `useSwipeEntryNavigation.tsx` — Hook that wraps content in a `PagerView` with prev/current/next pages. On swipe completion, navigates via `replaceControllerView` (same mechanism as pull-up-to-next).
- `EntryNavBar.tsx` — Prev/Next button bar with position counter. Buttons disable at list boundaries. Haptic feedback on navigation.

### Modified files
- `EntryDetailScreen.tsx` — Integrated swipe wrapper and nav bar. Extracted `navigateToEntry()` as shared function for all navigation methods.

### Architecture
- Uses the existing `entryIds` array already passed to `EntryDetailScreen`
- Uses `react-native-pager-view` (already a dependency) for gesture handling
- Follows the same `replaceControllerView` pattern as pull-up-to-next
- Pull-up-to-next continues to work alongside horizontal swipe
- Haptic feedback via `expo-haptics` (matching `PagerList` pattern)

## Test plan

- [ ] Open an article from a feed with multiple entries
- [ ] Swipe left → navigates to next article
- [ ] Swipe right → navigates to previous article
- [ ] Tap "Next ›" button → navigates to next article
- [ ] Tap "‹ Prev" button → navigates to previous article
- [ ] First article: "‹ Prev" is disabled/grayed
- [ ] Last article: "Next ›" is disabled/grayed
- [ ] Pull-up-to-next at bottom still works
- [ ] Position counter shows correct "N / Total"
- [ ] Vertical scrolling within article is not affected by horizontal swipe
- [ ] Test on iPhone and iPad

### DIFF
--- a/apps/mobile/src/modules/entry-content/swipe-entry-navigation/EntryNavBar.tsx
+++ b/apps/mobile/src/modules/entry-content/swipe-entry-navigation/EntryNavBar.tsx
@@ -1,0 +1,104 @@
+/**
+ * EntryNavBar — prev/next navigation buttons for article reading
+ *
+ * Shows at the bottom of the entry detail screen as a subtle bar.
+ * Provides tap targets for prev/next navigation (alternative to swipe).
+ * Also supports keyboard navigation on iPad (←→ arrows, J/K).
+ */
+
+import * as Haptics from "expo-haptics"
+import type { FC } from "react"
+import { useCallback } from "react"
+import { Pressable, StyleSheet, View } from "react-native"
+
+import { Text } from "@/src/components/ui/typography/Text"
+
+interface EntryNavBarProps {
+  prevEntryId: string | undefined
+  nextEntryId: string | undefined
+  onNavigateToEntry: (targetEntryId: string) => void
+  /** Current position in the list (1-indexed) */
+  currentIndex: number
+  /** Total entries in the list */
+  totalEntries: number
+}
+
+export const EntryNavBar: FC<EntryNavBarProps> = ({
+  prevEntryId,
+  nextEntryId,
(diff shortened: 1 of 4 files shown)


DIYgod:
Thank you for your pr, but I don't think horizontal swiping is a very good interaction because it conflicts with the right swipe exit operation.

PR #4940, opened 2026-03-31
fix(mobile): preserve pitch for Android video playback

luckycold:
### Description

Playback speed does not preserve the pitch of the video in certain contexts, specifically when tapping on a video directly from the feed. This fix just preserves the pitch after tapping on the video directly and increasing the playback speed to anything other than the normal playback speed.

### PR Type

- [ ] Feature
- [x] Bugfix
- [ ] Hotfix

Placed as video instead below.

https://github.com/user-attachments/assets/95c442e8-ff6a-4537-aa1f-501f0279a8af
> Video shows before and after the fix. The fixed version in the dev build comes at the 12 second mark.

### Linked Issues

None. Was just something I noticed that annoyed me, so I wanted to fix it.

### Changelog

### DIFF
--- a/apps/mobile/src/components/ui/video/VideoPlayer.tsx
+++ b/apps/mobile/src/components/ui/video/VideoPlayer.tsx
@@ -28,6 +28,7 @@ export function VideoPlayer({
   const player = useVideoPlayer(source, (player) => {
     player.loop = true
     player.muted = true
+    player.preservesPitch = true
     // player.play()
   })
   const { status } = useEvent(player, "statusChange", { status: player.status })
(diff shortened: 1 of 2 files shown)


luckycold:
@DIYgod I noticed it's merged and when I look on the Google Play store it says it's on 0.4.1. However, my playback speed fix isn't there. I tested with the test android build for the PR here, [and it works on this build your GitHub Actions created](https://github.com/RSSNext/Folo/actions/runs/23901641168). Not sure where the wires got crosses, but somehow the playback speed fix got undone along the way.

DIYgod:
Sorry, I forgot to add it to 0.4.1, I will include it in the next version.

PR #4984, opened 2026-04-20
fix: enter and return via the "discover" route

cuikaipeng:
### Description
when clicking into the discover page multiple times, the return also requires multiple clicks

### PR Type

- [ ] Feature
- [x] Bugfix
- [ ] Hotfix

https://github.com/user-attachments/assets/8647762c-4839-4bf9-b5be-5b03847b81c3

### Changelog


DIYgod:
Thank you

